PR #1, opened 2023-06-30
Add Magnetize pack

wavewhyse:


erasels:
Don't forget to add yourself to the author_list in the ModTheSpire.json!
Also, images are still in the works I assume?

src/main/java/thePackmaster/actions/magnetizepack/MagnetizeAction.java
```diff
@@ -0,0 +1,38 @@
...
+
+            for (AbstractPower pow : Wiz.adp().powers)
+                if (pow instanceof InduceCurrentPower)
+                    ((InduceCurrentPower) pow).onMagnetize(card);
+
+            if (card instanceof Ferroshield)
+                ((Ferroshield) card).onMagnetized();
+        }
+
+        CardModifierManager.addModifier(card, newMagnetize);
```
erasels, line 35: The modifier is still applied even if shouldApply is false? It's checked in addModifier, but since you already check it here, you might as well put it into the if clause

src/main/java/thePackmaster/cardmodifiers/magnetizepack/MagnetizedModifier.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private void spread(AbstractCard card) {
+        List<AbstractCard> candidates = AbstractDungeon.player.hand.group.stream().filter(c -> {
+            return c != card &&
+                    !CardModifierManager.hasModifier(c, ID) &&
+                    c.cost != -2;
+        }).collect(Collectors.toList());
+
+        if (candidates.size() > 0)
+            AbstractDungeon.actionManager.addToBottom(new MagnetizeAction(candidates.get(AbstractDungeon.cardRandomRng.random(candidates.size()-1))));
```
erasels, line 71: Might want to look at the Wiz class, it has a ton of qol methods that make things easier like:
`Wiz.atb(new MagnetizeAction(Wiz.getRandomItem(candidates));`

src/main/java/thePackmaster/cards/magnetizepack/Attraction.java
```diff
@@ -0,0 +1,44 @@
...
+    public Attraction() {
+        super(ID, 1, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.NONE);
+        baseMagicNumber = magicNumber = 3;
+        cardsToPreview = new Fuzz();
+    }
+
+    public void use(AbstractPlayer p, AbstractMonster m) {
+        for (int i = 0; i < magicNumber; i++) {
+            addToBot(new DrawCardAction(1));
+            if (p.hand.size() + i <= BaseMod.MAX_HAND_SIZE)
```
erasels, line 27: This approach could be mislead by things drawing inbetween these actions which could be done via beacon iirc or other modded relics

Might be a better idea to make into an anonymous action that checks if the hand isn't full. Basically jsut combine this check with the anonymous action you use to apply magentize

src/main/java/thePackmaster/cards/magnetizepack/PerfectedAlignment.java
```diff
@@ -0,0 +1,75 @@
...
+        count += Wiz.adp().drawPile   .group.stream().filter(card -> CardModifierManager.hasModifier(card, MagnetizedModifier.ID)).count();
+        count += Wiz.adp().discardPile.group.stream().filter(card -> CardModifierManager.hasModifier(card, MagnetizedModifier.ID)).count();
+
+//        return Stream.concat(Stream.concat(Wiz.adp().hand.group.stream(), Wiz.adp().drawPile.group.stream())Wiz.adp().discardPile.group.stream())
+//                .mapToInt(card -> CardModifierManager.hasModifier(card, MagnetizedModifier.ID) ? 1 : 0).sum();
+
+        return count;
+    }
+
+    public static boolean isStrike(AbstractCard c) {
```
erasels, line 47: unused?

src/main/resources/anniv5Resources/localization/eng/magnetizepack/Keywordstrings.json
```diff
@@ -0,0 +1,11 @@
+[
+  {
+    "PROPER_NAME": "Magnetized",
+    "NAMES": [
+      "Magnetized",
+      "magnetized",
+      "magnetize"
+    ],
+    "DESCRIPTION": "When you play or discard a #yMagnetized card, #yMagnetize another random card (excluding #yUnplayable cards) in your hand. Some cards have special effects with #yMagnetized cards."
```
erasels, line 9:
```suggestion
    "DESCRIPTION": "When you play or discard a #yMagnetized card, #yMagnetize another random card (excluding #yUnplayable cards) in your hand."
```
Bit too wordy, and it'll be self evident that it interacts with other things.

src/main/resources/anniv5Resources/localization/eng/magnetizepack/Powerstrings.json
```diff
@@ -0,0 +1,18 @@
...
+    "NAME": "Hyperpolarity",
+    "DESCRIPTIONS": ["The next #yMagnetized card you play is played twice.", "The next #b%d #yMagnetized cards you play are played twice."]
+  },
+  "${ModID}:InduceCurrentPower": {
+    "NAME": "Induce Current",
+    "DESCRIPTIONS": ["When you #yMagnetize a card, deal #b%d damage to a random enemy."]
+  },
+  "${ModID}:SuperconductancePower": {
+    "NAME": "Superconductance",
+    "DESCRIPTIONS": ["#yMagnetized cards deal #b%d more Attack damage and gain #b%d more Block."]
```
erasels, line 12:
```suggestion
    "DESCRIPTIONS": ["#yMagnetized cards deal #b%d more attack damage and gain #b%d more #yBlock."]
```

src/main/resources/anniv5Resources/localization/eng/magnetizepack/UIstrings.json
```diff
@@ -0,0 +1,24 @@
...
+  "${ModID}:MagnetizePackHat": {
+    "TEXT": [
+      "Always points north.",
+      "Magnet"
+    ]
+  },
+  "${ModID}:MagnetizedModifier": {
+    "TEXT": [
+      "Magnetized",
+      "When you play, discard, or Exhaust a #yMagnetized card, Magnetize another random card in your hand. Some cards have special effects with #yMagnetized cards."
```
erasels, line 18: Exhaust feature was removed
Also, you can remove the last sentence

src/main/java/thePackmaster/packs/MagnetizePack.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class MagnetizePack extends AbstractCardPack {
+    public static final String ID = SpireAnniversary5Mod.makeID(MagnetizePack.class.getSimpleName());
+    private static final UIStrings UI_STRINGS = CardCrawlGame.languagePack.getUIString(ID);
+    public static final String NAME = UI_STRINGS.TEXT[0];
+    public static final String DESC = UI_STRINGS.TEXT[1];
+    public static final String AUTHOR = UI_STRINGS.TEXT[2];
+
+    public MagnetizePack() {
+        super(ID, NAME, DESC, AUTHOR, new PackSummary(3, 2, 5, 2, 4, PackSummary.Tags.Exhaust, PackSummary.Tags.Tokens));
```
erasels, line 18:
```suggestion
        super(ID, NAME, DESC, AUTHOR, new PackSummary(3, 2, 5, 2, 4, PackSummary.Tags.Discard, PackSummary.Tags.Tokens));
```

src/main/java/thePackmaster/cards/magnetizepack/Attraction.java
```diff
@@ -0,0 +1,44 @@
...
+import com.megacrit.cardcrawl.actions.common.DrawCardAction;
+import com.megacrit.cardcrawl.actions.common.MakeTempCardInHandAction;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import thePackmaster.actions.magnetizepack.MagnetizeAction;
+import thePackmaster.cardmodifiers.magnetizepack.MagnetizedModifier;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class Attraction extends AbstractMagnetizeCard {
```
erasels, line 15: Actually, the draw action has a callback which you can use to apply your modifier, no need for an extra action:
You can look at Scrape from BaseGame or Rummage from PM for an example.

wavewhyse:
Yes, images are in the works. I did edit the json but it got lost in fiddling with Git.

src/main/resources/anniv5Resources/localization/eng/magnetizepack/Cardstrings.json
```diff
@@ -0,0 +1,48 @@
+{
```
erasels, line 1: It might be a good idea to have at least one discard card in here just to build the association with your mechanic

erasels (changes requested):

erasels:
I have some concerns for the cards in the pack, not necessairly balance, more in the sense that this pack might not feel very fun to play compared to packs that integrate better.

Magnetize doesn't really do anything on its own and doesn't synergize with anything outside of this pack, and even within this pack other than Ferroshield, all your synergy comes from synergy cards that do nothing outside of providing that synergy.
So, effectively, in a run you'd have to first pick a Magentized card and only then can you even look at the synergy cards, all the while you get a ton of cards from other packs that have an easier time fitting into your deck.

While that in of itself isn't an issue, we have a few inward-focused packs already, the current iteration of these cards are a bit... well, boring. I don't see myself being enticed to try them out without the synergy I would first have to draft into somehow. Packs that allow for more interaction with each other can be made to shine through these interactions, this pack can't rely on that.

I see that you wanted to allow for some of that by adding discard synergy, but you don't do anything with that either, sadly. There is no card that discard and no card that benfits from ebing discarded and no power that supports discard in some way either. 

Here are some suggestions for changes that I think could help with my above points, maybe they can give you some inspiration on how to ameliorate the current state:
Adding `Discard any number of cards.` to Superconductance.  I feel like this helps with supporting discard while also giving this card something to escape the weird flex vibe it's giving off right now.

Coilburst only adds Fuzz to you hand if it's not Magnetized.

I would honestly suggest completely replacing Hyper Polarity, it's decent but we have a strong coverage of this type of effect in packmaster to the point that this won't be able to turn heads. I think this slot might be better used for a card that brdiges the gap between your pack and others. 

Ohmic Push feels a little random now that Exhaust was removed from this pack. Another good candidate for a complete rework.

I think a power (as in an effect, not the card type) that makes the owner (an enemy) take damage when a card is Magnetized might be good. You could sprinkle that onto 2-3 cards (Coil burst could apply 2 of it for example) and you'd have a much smoother progression curve for your pack to be pickable due to preexisting synergy.
If you do go with taht route, Induce Current might clash too much with it, but that requires playtesting to say for sure.

wavewhyse:
>     * Whenever you play 10 Magentized cards, gain E.

I really like this idea, kinda does what I was trying to do with Ferroshield and World Dynamo and makes the support element of Magnetize be energy gain. Which frees up those card slots to do other things and yeah, incentivizes taking any magnetized card just for the keyword itself. I'm worried about how easily this might lead to infinites though, having it be a Nunchuk for any card (but you do have to scale into it). Especially with a card draw card in the pack, and free Fuzzes contributing to that 10 counter. Even so, I really want to implement it. What do you think?

erasels:
Yeah, I think it could work out well, we can always use some energy generaing packs as well, don't have too many of those. ANd it would free up some design space so we can get some more creativity in effects.

I don't really think you need to worry about infinites, 10 cards is still quite the hurdle. Something else would have to allow for a pseudo infinite already.

erasels:
Any updates on this?

wavewhyse:
hey, sorry. had to take a break from creating for a while. i did make updates to this but never pulled them or posted about it. do you want to check out the newest version?

erasels:
No problem, yeah, feel free to update the PR. PM is still accepting new packs

PR #12, opened 2023-09-23
Pick Them All pack

modargo:
This pack does a number of things that are unique for Packmaster, in a way that I find fun and interesting. I suspect there will be some people who just don't like the big deck theme and/or the inherent variance in some of the cards, but other than that this should be a pack people enjoy.

Here are the compendium versions of the cards, their updated versions, and their in-combat versions (Deck Check also gets text saying how much damage it deals even though it isn't shown in the image). The only change I made since this screenshot was increasing Touched by Flame's max HP loss by 1. Edit: and reducing the ignite gain from upgrading it by 1; the max HP cost is a significant enough drawback that it's okay for it to be a strong card, but in playtesting I found it was doing a bit too much.

pick them all pack cards
pick them all pack cards upgraded
pick them all pack cards in combat

General notes:
* The pickup effects are mostly positive: 6 positive (gaining cards/relics/gold), 1 mixed (Dark Bargain), and 3 negative. This assumes we could all the card gain as positive, but the idea is that it generally should be.
* There are 4 cards that directly synergize with big decks (Big Boots, Deck Check, Golden Pick, Pack Gear). The 3 rares provide indirect support while also working in various other deck types (Dark Bargain, Grab and Go, Transmogrifier). The remaining 3 cards don't have any particular synergy with big decks and help round out the pack.
* Many of the cards in this pack are a bit weak early on, though from my playtesting I found that it was completely fine to take a few cards from this pack early just to get some decent cards in your deck (anything is better than strikes and defends). Sometimes that leads into building a big deck, sometimes it doesn't, but the cards will still go up in value over the course of most runs. (Aggregate and Stack are good cards even at medium deck sizes, and if you take Deck Check it's usually because the other cards have some value.)
* Deck Check is really fun to see in card rewards. There are times where it isn't worth it, but there are also times where it feels great to take. Rare gems are excluded because getting the uncommon gems is already powerful enough.
* The rares are all very fun if you're willing to accept the variance! I think they're all reasonably strong cards, which is good because otherwise it can be hard to get people to take cards with downsides.
* There's more 6-line cards than I'd like, but I found it hard to do interesting pickup effects without some of the cards getting long. At least the in-combat versions are shorter.

Here's a fun example of what my deck was like at the end of act 1 in my most recent playtesting run:
pick them all pack lol deck

Here's the hat:
image


src/main/java/thePackmaster/cards/pickthemallpack/AbstractPickThemAllCard.java
```diff
@@ -0,0 +1,51 @@
...
+        this.rawDescription = cardStrings.DESCRIPTION + this.getPickupDescription();
+        this.initializeDescription();
+    }
+
+    @Override
+    public void upp() {
+    }
+
+    @Override
+    public final void uDesc() {
```
modargo, line 23: The purpose of all the code here (including in the constructor) is to allow separation between the base description and the pickup effect description, so that cards in this pack can automatically remove the pickup effect description during combat.

src/main/java/thePackmaster/packs/PickThemAllPack.java
```diff
@@ -0,0 +1,37 @@
...
+public class PickThemAllPack extends AbstractCardPack {
+    public static final String ID = SpireAnniversary5Mod.makeID("PickThemAllPack");
+    private static final UIStrings UI_STRINGS = CardCrawlGame.languagePack.getUIString(ID);
+    public static final String NAME = UI_STRINGS.TEXT[0];
+    public static final String DESC = UI_STRINGS.TEXT[1];
+    public static final String AUTHOR = UI_STRINGS.TEXT[2];
+    public static final String CREDITS = UI_STRINGS.TEXT[3];
+
+    public PickThemAllPack() {
+        super(ID, NAME, DESC, AUTHOR, CREDITS, new PackSummary(3, 3, 4, 3, 4));
```
modargo, line 19: This pack has a fair amount of variance, but calling it average offense/defense/frontload and above average support and scaling (various ways of improving your deck and building up big effects) seemed right from my playtesting.

src/main/java/thePackmaster/cards/pickthemallpack/BigBoots.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public void triggerOnGlowCheck() {
+        this.glowColor = this.checkPileSizes() ? GOLD_BORDER_GLOW_COLOR : BLUE_BORDER_GLOW_COLOR;
+    }
+
+    @Override
+    public void onObtainCard() {
+        AbstractDungeon.topLevelEffectsQueue.add(new ShowCardAndObtainEffect(new Stack(), Settings.WIDTH / 2.0f, Settings.HEIGHT / 2.0f));
```
modargo, line 54: Using the top level effects queue for this (and many of the other on obtain effects) makes animations display properly when the post-combat rewards screen is up.

src/main/java/thePackmaster/cards/pickthemallpack/BloodWall.java
```diff
@@ -0,0 +1,68 @@
...
+    public void onObtainCard() {
+        // We want to avoid negative scenarios like getting this from Astrolabe or Pandora's Box at low enough HP that
+        // it would kill the player, so we make this not reduce the player below 1 HP
+        int hpLoss = Math.min(PICKUP_HP_LOSS, AbstractDungeon.player.currentHealth);
+        if (hpLoss > 0) {
+            if (AbstractDungeon.getCurrRoom().phase == AbstractRoom.RoomPhase.COMBAT) {
+                this.addToBot(new DamageAction(AbstractDungeon.player, new DamageInfo(AbstractDungeon.player, hpLoss, DamageInfo.DamageType.HP_LOSS)));
+            }
+            else {
+                AbstractDungeon.topLevelEffectsQueue.add(new AbstractGameEffect() {
```
modargo, line 54: This needed to be put in an effect because of concurrent modification exceptions that would happen otherwise. I didn't track down the exact cause, but I assume the exceptions I saw were due to hooks or animations for taking damage.

src/main/resources/anniv5Resources/localization/eng/pickthemallpack/Cardstrings.json
```diff
@@ -0,0 +1,54 @@
...
+  "${ModID}:TreasureChest": {
+    "NAME": "Treasure Chest",
+    "DESCRIPTION": "Exhaust. NL Remove this from your deck after playing it !M! times.",
+    "EXTENDED_DESCRIPTION": [" NL ${ModID}:Pickup: Gain a random common relic."]
+  },
+  "${ModID}:GrabAndGo": {
+    "NAME": "Grab and Go",
+    "DESCRIPTION": "Unplayable. Innate.",
+    "UPGRADE_DESCRIPTION": "Unplayable.",
+    "EXTENDED_DESCRIPTION": [" NL ${ModID}:Pickup/Removed: Gain/lose a *Question Card. NL Remove this from your deck after the act boss.", "Unplayable. NL Innate."]
```
modargo, line 41: This wording is the best way I found to avoid very long card text.

Pickup/Removed is a variant of the pickup keyword that I added (in the main packmaster project, since that's where the keyword is defined, as part of the Bellord pack).

erasels:
Haven't looked at the code yet, just some thoughts regarding cards so far:
- Pack Gear could probably say Gain 1 + 1 for each metallicize instead of making it round up which runs contrary to all other rounding in the game iirc (would also be a little less text)
- Blood Wall might be a very unpopular card, while I understand the value of temp hp, sticking a small amount onto a defend and then taking HP to get it seems harsh
- Deck Check's a little too random for a common card, I think? What made you go from Ruby to random gem? That just feels like way too much variance to make an informed decision, although this would be stronger, I suppose.
- This is more of a personal thing, but I really dislike cards that are curses which you have to deal with to get meta rewards so Treasure Chest is a design that I really dislike. I do concede that there is some depth with trying to get it double played to reduce the 7/5 combat payoff (minimum)
- I don't really get Transmogifier. It will basically always hit a basic card (unless curses are considered lower) which is fine, but why does it also give you a strike? The card already seems pretty sub-par, why must there be a punishment for picking it?

I am not sure how good this Pack will end up being, when viewed in a vacuum it doesn't excite me very much. I can definitely see them being decent if you fully commit to them which may led to a bit of a bias in your testing runs. If you start out a run with the goal to make a thick deck to make this work, it'll have a good chance of making these cards pay off closer to the end. 
However, when not going into a run with the goal to make it work, you basically have to ask yourself whether this now sub-par card is good for a later pay-off which might not come if can't find a good avenue to go into a thick deck. 
Add to that, that a decent amount of these cards just aren't super pickable later into a run if you don't already have a thick deck and it might just end up being a pack that is hard to justify picking cards from at any point in the run.

I hesitate to give this kind of feedback since I haven't played it myself and I know that you're a way better player than myself, but these are my first reactions to see the draft. Although, I don't think i'm far from the avergae modded player.

modargo:
Thanks for the feedback! It got me to think about a few things that I think will result in nice improvements to the pack.

Tl;dr: tweaking Pack Gear/Deck Check/Transmogrifier, reworking/replacing Blood Wall/Treasure Chest, and IMO the other cards will be fine even if some people may shy away from taking the cards that seem like they require a big deck. Here are new screenshots, much longer responses with explanations of my thought process below (I like writing, and hopefully it's somewhat interesting to read). Haven't done any playtesting runs with these changes yet.

pick them all pack cards 2

pick them all pack cards upgraded 2

### Smaller points
* That's a good suggestion for Pack Gear, I'll change it.
* Blood Wall has been okay in testing, but weaker than I expected. It's only okay in block efficiency, so even with the nice benefit of temporary HP there are sometimes better defensive options. It's fine to take when you have HP to spare, but never really that good, and awkward if your HP is low and you don't have a way of healing. I was originally thinking of just buffing the block/temp HP numbers, but after thinking about it a bit more, I see an opportunity to redesign/rearrange things here -- the current effect of Blood Wall is weak enough that the HP loss could be replaced with a small upside instead.
* For Deck Check, the gem randomness is mostly to make it stronger, since Ruby is the weakest gem. I also like that it gives you interesting decisions to make. That said, it might be fine to settle on just one gem this always uses, just not Ruby. Maybe Tourmaline (the dice one) -- it's good both early and late, without being as strong as Onyx/Citrine (debuff removal/energy) or as deck-specific as Aquamarine (Frost). That does remove some of the joy of seeing an unexpectedly great combination, but reducing complexity is a real benefit.
* Fair point about Treasure Chest. Playing with it hasn't been very interesting, so I think I'll find a replacement for it. I can make something completely different and the pack will still have both Grab and Go and Transmutation as meta reward style cards.

### Blood Wall and Treasure Chest
Here's are my thoughts on Blood Wall and Treasure Chest.
* Rename Blood Wall to Bulk Up. Change it to Pickup: gain 1 Max HP, while leaving the block and temporary HP the same.
  * Pairing a small positive effect with a decent defensive card should make picking it feel better.
  * This is clearly a better card than Blood Wall. But I don't think it will be overpowered, since Blood Wall felt okay at beast.
  * This might be fine even with 2 Max HP. Regardless of the exact numbers, it's definitely possible to find block/temp HP/max HP values that make this balanced.
* Replace Treasure Chest with a new uncommon, Burning Fists. 1-cost Attack. Deal 4 damage twice. Gain 1 Strength. Pickup: lose 4 HP.
  * The idea is to have an attractive effect to overcome the HP loss, without being too powerful.
  * There's more than one option for the effect; I went with strength gain because its good and simple. It gives this pack an alternate way to contribute to a variety of deck types and is a card that should be worth taking early.
  * Repeatable strength gain should be fine on an uncommon; it will be good when you're able to pick up other multi-attacks, but at only 1 strength per play it's pretty slow to ramp up.
  * If we were worried about this card we could give it Exhaustive, but I'd prefer to find values for damage and HP loss that seem balanced.

### Transmogrifier
Here's how I would think about Transmogrifier: it's an incremental Pandora's Box that can clean out more cards after your Strikes and Defends (Rummage is often bad enough you want to remove it in the late game, and often you have bad commons you took early). Pandora's Box is one of the strongest boss relics, so one way to evaluate this card is to ask what price you'd pay for an extra boss relic. We can say the price for this card is two curses (itself and the Strike), which isn't crazy as a baseline.

Getting two dead cards in your deck is less bad than it seems, since with your basics handled by Transmogrifier, you often have removals to spare to get rid of the Strike and eventually Transmogrifier itself. So it's about how much you can afford the incremental aspect and benefit from transforming extra cards that Pandora's Box wouldn't. In my playtesting, I've taken this when I've had 2+ Strikes left in my deck, and I've found that my deck was usually better after 2-3 fights because Strikes get transformed first.

The other issue to consider is that I've seen a lot of players say that they think Pandora's Box is bad or that they just don't like it. That seems to me like the core dilemma here: if you have something that's fairly strong but that some players don't like, what do you do? There are options, but it's tricky to find one that satisfies everyone. Pandora's Box is an example of a case where the devs seem to have decided that it was fine as is.

As a mod, Packmaster has more leeway on balance than the base game, so my thought is to give Transmogrifier a slight buff and have it add a random common instead of a Strike. That's definitely better, but it's not so much better that it wrecks the balance of the card. It's also very on theme for the card.

### Overall play experience of the pack
For the broader feedback about this pack, I don't see picking these cards as being about committing to a big deck early on. It's more about being open to that as a possibility, and flexibly evaluating picks as they come based on what challenges you're facing next. Of course, I think that's already different from how many people play the game; if someone is coming in with preconceptions of what archetypes or style or deck size they will/won't play, or they feel committed to a specific strategy after a few picks, or they don't like cards with variance, this pack will be less attractive.

In a way, this is the Pandora's Box dilemma extended to all the high variance/big deck cards in the pack, and it's hard to work out a comprehensive solution. I like how things play enough from my playtesting that I don't want to redesign these. Instead, perhaps buffing some of the cards is an option; I've thought of ways to do it for all the big deck cards. Deck Check could could always cost 1 and have the upgrade add "+ 3" like how Stack does it, and Big Boots/Pack Gear could get simple numeric adjustments.

But while I was fine with a buff to Transmogrifier since it's a rare you don't want multiple copies of, I'm wary of too many buffs to the commons and uncommons. Doing that can make it too easy to quickly get above the power curve in runs, and I think all the big deck cards are reasonable picks throughout the game for a decent range of decks. That belief is reinforced by seeing from the metrics thread that the average winning deck size for Packmaster is 29 cards. That's a size where the big deck cards are definitely good, and where they'd be worth taking at many points during the run (especially since having this pack in the run encourages ending with a higher than average deck size).

Doesn't mean I'm completely ruling out any buffs or other changes, just that my first hope is that the changes to the other cards will let this pack be reasonably attractive to the average modded player. Happy to continue discussing and take suggestions.

erasels:
Oh, I didn't make the connection that the usage of deck in Transmorifier means that it's permanent. I was wondering why you didn't say drawpile and figured it was just because Grave cards exist.

modargo:
> Oh, I didn't make the connection that the usage of deck in Transmorifier means that it's permanent. I was wondering why you didn't say drawpile and figured it was just because Grave cards exist.
Ah, I see. Maybe I can fit the word "permanently" on the card if I saved some text length by changing it back to adding Strike. Think I should?

> I like the changes you made, definitely makes Bulk Up more attractive. Not sure when exactly I'll get around to the code review, but if I do end up making a pack myself, I'll likely wait with the update until my pack is ready as well.
Sounds good. If you decide to make one, I look forward to seeing it!

erasels:
> Think I should?
Worth a try to see how it looks.

modargo:
Looks like it fits, so I think I'll go with this version. The text spacing is really awkward due to the long words, but clarifying that it's permanent seems worth it to me.

image

modargo:
Made a few balance changes:
* Reduced Touched by Flame upgrade ignite by 1. Even with having reduced this already, the upgrade still felt too high value. I'm confident the card is still quite strong (which it should be, given the max HP cost).
* Changed Bulk Up to give one less block and one more temporary HP. It felt like the card could use a small buff.
* Changed Burning Fists to do one hit of 8 damage instead of two hits of 4 damage. While I don't think it was overpowered, with hitting twice it felt a bit too easy to take advantage of the strength gain. With how many multi-hit attacks there are in Packmaster, the recurring strength should still get people to view this as a card worth taking.

src/main/java/thePackmaster/cards/pickthemallpack/BigBoots.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final int COST = 1;
+    private static final int DRAW = 2;
+    private static final int UPGRADE_DRAW = 1;
+    private static final int THRESHOLD = 15;
+
+    public BigBoots() {
+        super(ID, COST, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.SELF);
+        this.magicNumber = this.baseMagicNumber = DRAW;
+        this.secondMagic = this.baseSecondMagic = THRESHOLD;
+        this.cardsToPreview = new Stack();
```
erasels, line 25: Might be too much effort, but I don#t see any logic to hid this card preview when in combat. Not sure if it needs to exist once the card is in the deck at all.

erasels (changes requested):

src/main/java/thePackmaster/cards/pickthemallpack/BigBoots.java
modargo: That logic exists, it's in `AbstractPickThemAllCard.resetDescriptionForCombat`, because all card previews in this pack are for pickup effects so they can all be removed for combat.

src/main/java/thePackmaster/cards/pickthemallpack/BurningFists.java
```diff
@@ -0,0 +1,67 @@
...
+        this.addToBot(new ApplyPowerAction(p, p, new StrengthPower(p, this.magicNumber)));
+    }
+
+    @Override
+    public void onObtainCard() {
+        // We want to avoid negative scenarios like getting this from Astrolabe or Pandora's Box at low enough HP that
+        // it would kill the player, so we make this not reduce the player below 1 HP
+        int hpLoss = Math.min(PICKUP_HP_LOSS, AbstractDungeon.player.currentHealth - 1);
+        if (hpLoss > 0) {
+            if (AbstractDungeon.getCurrRoom().phase == AbstractRoom.RoomPhase.COMBAT) {
```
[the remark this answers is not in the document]
modargo: I tested the Neow scenario -- you actually have a room even at Neow, check out `NeowRoom`.

Still, there's enough other odd stuff that could happen that using the Wiz function is safer, so I'll switch over to that.

erasels:
Good job as always

PR #25, opened 2023-11-10
Preliminary Disco Pack consideration

NoMasPollo:
Disco Pack Display
This is something i've been working on for a while. Obviously, its not done, the powers, cards, and pack dont have art yet but I do have someone I can ask to help me with that. This is more of asking if its balanced or too busted. I did add a secondBlock since initially I was using shenanigans with magic number but they would sometimes update strangely. If you know a better way of doing it please let me know bc i only use it once.


src/main/java/thePackmaster/actions/discopack/SpecificToHandFromDiscardAction.java
```diff
@@ -0,0 +1,39 @@
...
+import com.megacrit.cardcrawl.actions.AbstractGameAction;
+import com.megacrit.cardcrawl.cards.AbstractCard;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import thePackmaster.SpireAnniversary5Mod;
+
+import java.util.ArrayList;
+
+public class SpecificToHandFromDiscardAction extends AbstractGameAction {
```
erasels, line 14: `BetterDiscardPileToHandAction`
Exists, this is not needed

src/main/java/thePackmaster/actions/discopack/AnyNumberDiscardAction.java
```diff
@@ -0,0 +1,39 @@
...
+import static thePackmaster.util.Wiz.p;
+
+public class AnyNumberDiscardAction extends AbstractGameAction {
+
+    private final List<AbstractCard> selectedCards = new ArrayList<>();
+
+    public AnyNumberDiscardAction(List<AbstractCard> selectedCards) {
+        this.amount = amount;
+        this.selectedCards.addAll(selectedCards);
+    }
```
erasels, line 23: This seems unneeded, just give the SelectCardsInHand action a a lamda with a for each loop that adds discardspecificcardaction on them.

That also will contain any patches made to the discard action.

src/main/java/thePackmaster/vfx/discopack/SpotlightFXtest.java
```diff
@@ -0,0 +1,57 @@
...
+import com.megacrit.cardcrawl.core.AbstractCreature;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.helpers.ImageMaster;
+import com.megacrit.cardcrawl.vfx.AbstractGameEffect;
+import thePackmaster.SpireAnniversary5Mod;
+
+import static thePackmaster.SpireAnniversary5Mod.modID;
+
+public class SpotlightFXtest extends AbstractGameEffect {
```
erasels, line 17: Might be nice to rename this, test seems kind of weird here since it is being used

src/main/resources/anniv5Resources/localization/eng/discopack/Cardstrings.json
```diff
@@ -0,0 +1,46 @@
...
+    "DESCRIPTION": "Draw !M! cards. Discard a card. NL ${ModID}:Smooth: Draw !${ModID}:m2! cards."
+  },
+  "${ModID}:InTheRhythm": {
+    "NAME": "In The Rhythm",
+    "DESCRIPTION": "Whenever you discard !${ModID}:m2! cards, gain !M! Strength and 4 stslib:Temporary_HP.",
+    "UPGRADE_DESCRIPTION":"Innate. NL Whenever you discard !${ModID}:m2! cards, gain !M! Strength and 4 stslib:Temporary_HP."
+  },
+  "${ModID}:PointMove": {
+    "NAME": "Point Move",
+    "DESCRIPTION": "Place a card from your discard pile on top of your deck. NL ${ModID}:Smooth: Gain [E] .",
```
erasels, line 26:
```suggestion
    "DESCRIPTION": "Place a card from your discard pile on top of your drawpile. NL ${ModID}:Smooth: Gain [E] .",
```

src/main/resources/anniv5Resources/localization/eng/discopack/Cardstrings.json
```diff
@@ -0,0 +1,46 @@
...
+    "UPGRADE_DESCRIPTION": "Draw a card, then place a card from your discard pile on top of your deck. NL ${ModID}:Smooth: Gain [E] ."
```
erasels, line 27:
```suggestion
    "UPGRADE_DESCRIPTION": "Draw a card, then place a card from your discard pile on top of your drawpile. NL ${ModID}:Smooth: Gain [E] ."
```

src/main/resources/anniv5Resources/localization/eng/discopack/Cardstrings.json
```diff
@@ -0,0 +1,46 @@
...
+    "NAME": "The Bump",
+    "DESCRIPTION": "Deal !D! damage. NL Discard a card. NL ${ModID}:Smooth: Deal !${ModID}:sd! damage to a random enemy."
+  },
+  "${ModID}:TheBusStop": {
+    "NAME": "The Bus Stop",
+    "DESCRIPTION": "Gain !B! Block. NL Discard a card. NL ${ModID}:Smooth: Gain !${ModID}:sb! Block."
+  },
+  "${ModID}:TheRobot": {
+    "NAME": "The Robot",
+    "DESCRIPTION": "Deal !D! damage. NL Until next turn, whenever you discard a card gain !M! Block."
```
erasels, line 39:
```suggestion
    "DESCRIPTION": "Deal !D! damage. NL Until the end of your turn, whenever you discard a card gain !M! Block."
```

erasels (changes requested):
Still formulating my thoughts on pack design and balance, here's just some small thigns I noticed in the code

NoMasPollo:
Honestly, I agree with most of the points here. Halfway through making the pack, I realized how actually hard it is to make a non-insular discard pack. A number of the cards went through a lot of changes to make them feel less bad (and funky chicken used to be an acro clone until i realized that it was literally just acro.) Some of the extremely recent changes were hard nerfs to a lot of the cards. I agree that some of them are a little bland, I think I was worried about complexity for the sake of complexity in how they worked. A big question is any restrictions on using other pack's gimmicks, I was thinking about using inspiration seeing as bard and this are both flavored towards music, or maybe making fast movements a flurry card as one of those powers where it really matters when you play it. Point move is intentionally rare infinite bait, though it originally just put the card into your hand. Is a rare 0e hologram without the block but with other upside op? Idk. One of the big problems was deciding what the commons did, especially the really basic ones. I might keep just one tbh, probably the block one. I'm gonna make a mockup of what this theoretic rework would be, now that I think of some alternatives.

1: Yeah, that's never actually used. I don't remember why I even put it in anymore.
2: Whoops, sorry I forgot to rename it from test since I haven't been working with it for a bit. Will fix.

NoMasPollo:
disco display 2
Here's a potential rework I was looking at. Numbers subject to change. A lot of these have either used some mechanics from other packs or the original (more complex xd) concepts.

erasels:
Oh, I like the idea of including some Bard flavor via inspire.

Some notes:
Funky Chicken being 0 cost draw might still be an issue, infinites do not care about next turn.
Electric Slide: I don#t think your pack really needs to dabble in orbs, doesn't really fit. You've got eneough going on when just focusing on the basics around card manipulation.
Point Move, please go away from 0 cost draw card.
Fast movements will definitely have to be double checked value wise, it's basically the Bard's Form card if you have no cards in hand.
Explosive Finish: I don't think locking fits here and it adds a bit more complexity than the card should have in my opinion.


I don't think the pack is quite there yet, it doesn't really cleany integrate with other packs at the moment and the discard is still too heavily relied on. I liked the change to Fast movements, a nice way to implicity encourage discard but also allowing other archetypes to shine.
I'd say try to go for 3 cards that don't explicity mention discard at all.

Now that you've added inspiration your deck has a bit more identity in the role it tries to slot into. Since inspiration is easy to lose (by playing any attack ro block card) timing matters a lot, so maybe try adding a card that helps with it.
My thinking here is that inspiration likes big cards (so stuff like YMCA benfits the least from it counterintuitively) so it necessitates bigger cards being in hand when needed. Now how to mix that into discard...

NoMasPollo:
image
New concepts. One of the ones I am hesitant to change is YMCA and being tied to times, mostly because I already coded the vfx for it and it would cause me great pain to change that xd. This version takes advantage of how the amount2 of inspiration works. Note that there are a few rarity and type changes.

erasels:
Much better! This is starting to look great, some value tweaks and one more card rework and we should have it.

The Bump's block needs to be decreased a bit since it's basically free Block and upgrading both values by 4 is a bit much. I can see a 2 for each, or 4 for one. Or maybe a 3/1 split.

Point Move needs to go back the drawing board, it's fine if you want it to keep being a damage card, but right now it's a bit bland and also *very* slow. Compare it to rampage's values and even that is considered a bad card because it's too slow. It's a 11 damage card at base which isn't bad, but there's no real way to scale it and being discard fodder does not make for an interesting card.

Fast movements, as mentioned before, might have to be reduced to 25/50 inspiration or 3/2 cost with 50 constant inspiration. Or maybe changing the threshold to 2, a lot of avenues to tweak it so i'm not worried.

Explosive Finish looks interesting, playtesting will have to reveal if it's too easy to trigger for its large numbers or not. Wording wise, i'd probably remove the ", then". It might need a tune down on the numbers, we'll have to see.

YMCA's idea is fine, but right now it's a bit complex and hard to parse. I'd say limit it to gaining inspiration more times and increase the base gain by 5. We'll have to see if that feels like it scales too slowly or not, but I think it will be more healthy or the card as a whole even if the current idea is very flavorful.

NoMasPollo:
image

It's really hard coming up with that last card tbh. This should hopefully help make it feel better for scaling, as it gives "free" discard fodder as well as having a big damage hit. The idea with Fast Movements now that its rare, is that its a conditional sidegraded Splendid Form. If your hand is empty, it gives a 25 - 50 larger total % boost to cards, but the strength of splendid form is that it gives you all that value on one card which makes high values stronger. YMCA is now just farming 25% inspiration as much as possible to get a lot of value.

erasels:
What do you think about making Ste Tap ethereal, so you have to discard it (implicit synergy, yay) or play it to continue to be able tot let it generate a decent 2 cost?

NoMasPollo:
1: Sounds interesting, will definitely make it scale less though
2: Im not particularly attached to the name, but the idea was that it "dances" from your hand to your discard and back. Something along those lines would be fine.

erasels:
@NoMasPollo no pressure since I'll start merging stuff in Febuary again, just curious if this pack is still in the works? I thought it was pretty cool and just want to make sure it's still on the radar :)

NoMasPollo:
Yeah, sorry I've had school but I'm working on getting art

PR #28, opened 2023-11-20
Pick Them All pack: prevent Transmogrifier from transforming cards that cannot normally be removed from your deck

modargo:


erasels:
Oh, I thought the card skipped curses, well, I guess there are other card types as well that are soulbound

modargo:
> Oh, I thought the card skipped curses, well, I guess there are other card types as well that are soulbound

Yeah, normal curses didn't have any issues because they use the Curse rarity. The issues would come up with (1) curses with other rarities (e.g. Ascender's Bane/Necronomicurse/Curse of the Ball are all Special rarity) and (2) modded Soulbound cards.

I didn't actually run into this in gameplay; just noticed it while looking at the code and then confirmed my hypothesis in game. It takes a _really_ niche situation to actually trigger this, because Special rarity is treated the same as Rare, so your whole deck would need to be Rare or Special before a Special rarity card could be transformed.

PR #29, opened 2023-11-24
- Intrigue Pack: Reloaded

AlexMdle:
Reworked a substantial amount of the pack to be more interesting/intuitive/useful. Polished up the card art to match.


src/main/java/thePackmaster/patches/intriguepack/ExultationPatch.java
```diff
@@ -0,0 +1,58 @@
...
+import thePackmaster.cards.intriguepack.Exultation;
+
+public class ExultationPatch {
+
+    @SpirePatch(
+            clz = FastCardObtainEffect.class,
+            method = "update"
+    )
+    public static class OnPickupCardDoStuffPatch2 {
+        public OnPickupCardDoStuffPatch2() {
```
modargo, line 18: Very minor, but you can delete this constructor -- it's not doing anything so it's not needed.

Same for the constructor of the other patch below.

src/main/java/thePackmaster/cards/intriguepack/Favor.java
```diff
@@ -24,18 +20,14 @@ public Favor() {
...
-        if (eligibleCards.size() > 0) {
-            AbstractCard cardy = eligibleCards.get(AbstractDungeon.cardRandomRng.random(0, eligibleCards.size() - 1)).makeCopy();
-            cardy.setCostForTurn(0);
-            CardModifierManager.addModifier(cardy, new ExhaustMod());
-            CardModifierManager.addModifier(cardy, new EchoedEtherealMod());
-            CardModifierManager.addModifier(cardy, new GlowEchoMod());
-            Wiz.atb(new MakeTempCardInHandAction(cardy, true));
-        }
+        // Wow, so much easier once I found this lmao.
+        CardGroup cards = JediUtil.filterCardsForDiscovery(c -> !c.hasTag(CardTags.HEALING) && c.rarity == CardRarity.RARE);
```
modargo, line 24: This has a difference that I don't think you want -- it gets rares of all colors and packs, which isn't how the previous code worked.

Here's what you can do to only get the right cards:
```suggestion
        List<AbstractCard> eligibleCards = Wiz.getCardsMatchingPredicate(c -> c.rarity == CardRarity.RARE && !c.hasTag(AbstractCard.CardTags.HEALING));
        CardGroup cards = new CardGroup(CardGroup.CardGroupType.UNSPECIFIED);
        cards.group.addAll(eligibleCards);
```

A few more lines, but not that bad.

modargo (commented):
Looking good, this will be a nice boost to the pack.

Only one real issue that I saw (with Favor letting you find cards from other characters). Other comments are just code cleanup.

AlexMdle:
Fixes up.

modargo:
Merged! This will go out with the next release, which will probably be in early December.

PR #32, opened 2023-11-29
 Downfall pack: change Awaken Death to trigger before Fairy in a Bottle or Lizard Tail

modargo:
Leaving this up for you to merge in case you want to wait longer or try again with getting this fix in StsLib.

You can also take a quick look over things and read the comments I wrote. The patch location I picked is easy to target, but it's inside a block of code that doesn't run if you have Mark of the Bloom. Essentially, it assumes on death powers are healing, which is accurate for the one and only power like this in the mod. Not something we need to worry about, but maybe Kio would care about other use cases for this interface and want to target a different location.


erasels:
Yeah I was thinking about this, I don't think we'll be getting any updates out of kio before the anniversary project is close to wrapping up so I'm happy with having this in the mod.

PR #37, opened 2023-12-12
Graveyard pack: change Grey Bargain to give the card it returns purge

modargo:
Leaving this one up for you to merge in case you want to look over it or discuss further.

The PurgeModifier class that I repurposed doesn't seem to be used by anything. I assume it was used at some point, but the card that used it got reworked. Regardless, it was convenient for me since it meant the text was already there, so I just swapped it over to using the new PurgeField.

### DIFF
--- a/src/main/java/thePackmaster/cardmodifiers/cosmoscommand/PurgeModifier.java
+++ b/src/main/java/thePackmaster/cardmodifiers/cosmoscommand/PurgeModifier.java
@@ -1,6 +1,7 @@
 package thePackmaster.cardmodifiers.cosmoscommand;
 
 import basemod.abstracts.AbstractCardModifier;
+import com.evacipated.cardcrawl.mod.stslib.fields.cards.AbstractCard.PurgeField;
 import com.megacrit.cardcrawl.cards.AbstractCard;
 import com.megacrit.cardcrawl.core.CardCrawlGame;
 import com.megacrit.cardcrawl.localization.UIStrings;
@@ -8,7 +9,7 @@ import com.megacrit.cardcrawl.localization.UIStrings;
 import static thePackmaster.SpireAnniversary5Mod.makeID;
 
 public class PurgeModifier extends AbstractCardModifier {
-    public static String ID = makeID("PurgeCardModifier");
+    public static String ID = makeID("PurgeModifier");
     private static final UIStrings uiStrings = CardCrawlGame.languagePack.getUIString(makeID("PurgeModifier"));
     public static final String[] TEXT = uiStrings.TEXT;
 
@@ -19,15 +20,15 @@ public class PurgeModifier extends AbstractCardModifier {
     }
 
     public boolean shouldApply(AbstractCard card) {
-        return !card.purgeOnUse;
+        return !PurgeField.purge.get(card);
     }
 
     public void onInitialApplication(AbstractCard card) {
-        card.purgeOnUse = true;
+        PurgeField.purge.set(card, true);
     }
 
     public void onRemove(AbstractCard card) {
-        card.purgeOnUse = false;
+        PurgeField.purge.set(card, false);
     }
 
     public AbstractCardModifier makeCopy() {
--- a/src/main/java/thePackmaster/cards/graveyardpack/AncientSpear.java
+++ b/src/main/java/thePackmaster/cards/graveyardpack/AncientSpear.java
@@ -26,7 +26,7 @@ public class AncientSpear
   public AncientSpear() {
 	super(ID, COST, TYPE, RARITY, TARGET);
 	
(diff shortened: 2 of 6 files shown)


erasels:
Totally fine with this change but the pack itself suffers a good bit for it.
Looking at the cards: https://slay.ocean.lol/w/the%20packmaster%20expansion%20packs/anniv5%3agraveyardpack
Most fo them are fine cards that exhaust when used, however Extra Limbs kind of depends on being exhumed.

Looking at the pack, I think it's in a pretty bad spot. Other than the 3 exhumes and the burst-like, it's in a pretty bad shape.

modargo:
> Totally fine with this change but the pack itself suffers a good bit for it. Looking at the cards: https://slay.ocean.lol/w/the%20packmaster%20expansion%20packs/anniv5%3agraveyardpack Most fo them are fine cards that exhaust when used, however Extra Limbs kind of depends on being exhumed.
> 
> Looking at the pack, I think it's in a pretty bad spot. Other than the 3 exhumes and the burst-like, it consists of alright cards that exhaust. With its main way of getting those back also purging them, most of them seem like they won't be worth the effort.

I agree. My feeling about the Graveyard pack is that it's a mix of potentially degenerate effects and cards that are simply too weak. The sad thing about Extra Limbs and the other exhaust cards is that even without this change to Grey Bagain, there are normally better cards to get back (and better cards to put in your deck in the first place). I can see that the design intent was probably that the pack shouldn't provide too strong of a payoff for its own effects... but it went too far and made all the payoffs too weak.

Here's my assessment of the exhaust cards:
* Extra Limbs is okay, since 12 damage is a good rate for a 1-cost common attack. Scaling by getting it back multiple times has never felt very realistic to me (and there are usually better ways to scale, like applying poison), but you can take it early.
* Ancient Spear is a contender for the most boring card in the mod. It's also the exact same unupgraded as https://slay.ocean.lol/w/the%20packmaster/anniv5:firevial (the upgrades are different).
* Flesh Like Oak has numbers that are just too low for an uncommon that exhausts. It's barely more efficient than Defends, and the 1 strength isn't enough of a benefit to compensate. I mean, https://slay.ocean.lol/w/the%20packmaster/anniv5:titansstrength exists and is strictly better, and that card's win rate is good but not great (34.87%, around 130 out of 400 uncommons).
* Funerary Jewels is okay because it's net neutral on energy.
* Spontaneous Ritual can do powerful things but usually scales slower than other effects like this such as  https://slay.ocean.lol/w/the%20packmaster/anniv5:overtime or https://slay.ocean.lol/w/the%20packmaster/anniv5:dragonmeteor. And while those cards are rare, they're not even always good, since there are often better scaling payoffs in your card pool.
* Tombstone is only useful with Snecko Eye and pales in comparison to other 3-cost cards, e.g. https://slay.ocean.lol/w/the%20packmaster/anniv5:shifting (just to pick one comparison that comes to mind).

If you're counting, that's every card that isn't doing buried card stuff or exhaust synergy. In summary, the pack has four potentially degenerate cards (three of which will now have Purge), two okay cards, and four weak cards. Not the most exciting pack.

There's an obvious place this goes for me -- how about we buff the weak cards? Flesh Like Oak could get +1 block and +1 strength, Tombstone could get +5 block, Ancient Spear could get +2 damage. Not sure how to change Spontaneous Ritual, and I'm inclined to leave Extra Limbs and Funerary Jewels alone for now. That seems like the most straightforward way to improve the pack without trying to redesign it (since a lot of the issues are inherent to the pack's core design).

The funny thing is that I remember suggesting buffs like this during development. Better late than never I suppose.

erasels:
Hm, fair point regarding Extra Limbs. It's not actually bad, even if the scaling on it seems a bit superfluous at this point. (getting 12 and 18 damage out of it with double/triple str scaling)
Spontaneous Ritual will probably just be a very low pickrate card but can definitely be powerful for Distortion or other exhaust payoff decks. It's probably fine like that.

Other buffs seem fine to me, most of the cards are still pretty boring but at least they won't be dead in the water. Will you be adding those changes to this PR?

modargo:
> Hm, fair point regarding Extra Limbs. It's not actually bad, even if the scaling on it seems a bit superfluous at this point. (getting 12 and 18 damage out of it with double/triple str scaling) Spontaneous Ritual will probably just be a very low pickrate card but can definitely be powerful for Distortion or other exhaust payoff decks. It's probably fine like that. If it really needs a buff we could decrease the cost to 0 to make it easier to get the good cards out of your hand before you play it.
> 
> Other buffs seem fine to me, most of the cards are still pretty boring but at least they won't be dead in the water. Will you be adding those changes to this PR?

Yeah, I'll add them to this PR and merge it, since we're in agreement on the changes here and they're completely straightforward.

PR #39, opened 2023-12-23
Adds Royalty Pack, probably the winner of kiooeht's "ultra prestigious, definitely real, award of 'last pack added before anniv6'" v_v

Clauvin:
AT LEAST \O/

This PR adds the Royalty Pack to Packmaster's Expansion Pack - an adaptation of the Royalty class from Buriedbornes.

It's cards are strong but have costs varying from "discard one random card" to "weak but you can spend gold or HP for better effects". Also it has a hat 👍 , and an initial summary.

Given both the unusual use of gold as part of cost to play the cards, and the time until next February for a proper release, I will be testing the pack through January to catch possible needed buffs/nerfs (_I'm counting that Throw Soulstones definitively will need a nerf of some sort, just can't put my finger on which one precisely_), so maybe I will do another PR before February.


Clauvin:
Closing this because after doing some playtesting (and after some good points from Vex, Shammy and modargo at the Discord server), I'm rebalancing the pack AND opening it for playtest before committing it again.

PR #84, opened 2025-01-29
Into the Breach pack rework

ActuallyEnbeon:
itbrework
### Summary of changes
- Removed Electric Whip, Enrage Shot, and Storm Generator
- Added Shield Projector, Field Research, and Flamethrower
- Demoted Smoke Pellets to common ~~and reduced its Block gain by 1~~
- Added a visual effect to monsters with A.C.I.D.

Please advise on whether the updated pack ratings (4, 3, 3, 3, 2; Debuffs tag) are suitable


modargo (approved):
The overall changes look good. Values for the new cards seem reasonably well balanced and the rares are interesting. I made one balance suggestion that you can decide on.

The pack definitely still has more offense than defense and more frontload than scaling -- but the scaling is definitely improved a bit -- so the new ratings seem like a good start. It's easy enough to adjust them later if playing with the pack makes us think differently about it.

PR #87, opened 2025-02-06
Blade Storm pack

JDodi:
This PR adds the Blade Storm pack to Packmaster, supporting the heavy use of attacks in decks with on-the-go protection and damage scaling.

EDIT: Compendium, up to date as of Feb 13 2025 :

BladeStormPack Compendium
BladeStormPack Compendium (U)


JDodi:
BladeStormPack Compendium
BladeStormPack Compendium (U)
The packs's cards in the Compendium.

JDodi:
BladeStormPack Compendium
BladeStormPack Compendium (U)
The pack's cards in the Compendium (fixed)

erasels:
Modargo is going to do code review, so I'm just going to give some opinions on cards:
- Generally very unhappy with your text formatting and wording. Use NLs for effects that are separate from one another and if that makes the text hard to read then it's probably a good idea to rethink the wording or polish the effect a bit
- Overpressure makes it seem like the first effect would be affected by frail. 
- Not sure if the pseudo-Fatal effect on Fallover does much
- Tempest of Strikes is too many hits, especially with the availability of strength

JDodi:
Thanks for the feedback. 

- I'll work on the wording and formatting.
- In Overpressure, the "First" text was initially at the beginning, which solves that impression about Frail, but then I was told (or realized) that "First" and "Startup" text always comes after the normal text, which led to this. So it's ok to put it at the beginning again?
- If I make FallOver's Bind lower, on a random other enemy (so it still can't happen on the target) and not based on Fatal, is that good? When it kills, this will result in the same effect, just weaker.
- I was thinking about making Tempest of Strikes divide the number of attacks by 2 and raising the effect per attack. How does "6 damage and 2 Bind per 2 attacks in your deck" sound? I think it can handle the reduced synergy with Strength

modargo:
I'll also give the wording a close look when doing the code review.

For Overpressure, it's probably worth breaking convention by putting the `First` effect before the other effects. Otherwise a lot of people will wonder the same thing Gk did (whether the Frail affects the second instance of block). I'd judge that confusion over what is or isn't part of the `First` effect will be less of an issue. When there are two separate "Gain 7 Block" sentences, with a line break between them, there's only one sensible interpretation.

JDodi:
BladeStormPack Compendium
BladeStormPack Compendium (U)
Better? Strictly base game or PM wordings if I'm not mistaken, except where there was no precedent, in which case i stitched together base game and/or PM wordings.
Also simplified the pseudo-fatal on FallOver, and the upgrade on Tempest of Strikes since there wasn't space anymore for the old upgrade's text.
This isn't in the PR yet as I'm awaiting confirmation.

JDodi:
Actually never mind "Fall Over", I put the Strength gain at the top thinking that not putting it between the damage and what the damage can cause was better for clarity. But that messes with calculating damage since Strength would influence it. So I'm going to put that line at the bottom instead.

JDodi:
Or is it better to change "Fall Over" further? It's supposed to be a damage scaling tool, that can be sacrificed instead for a short respite in fights with a lot of incoming damage. So these versions would also fit, if you like them better :
- "[Attack] Deal 7 (10) damage. Apply 6 Bind to ALL enemies. When this is drawn, gain 1 Strength. Exhaust."
- "[Power] At the end of your turn, if you have killed an enemy this turn, apply 10 (15) Bind to a random enemy, otherwise gain 1 Strength."

modargo:
> Or is it better to change "Fall Over" further? It's supposed to be a damage scaling tool, that can be sacrificed instead for a short respite in fights with a lot of incoming damage. So these versions would also fit, if you like them better :
> 
> * "[Attack] Deal 7 (10) damage. Apply 6 Bind to ALL enemies. When this is drawn, gain 1 Strength. Exhaust."
> * "[Power] At the end of your turn, if you have killed an enemy this turn, apply 10 (15) Bind to a random enemy, otherwise gain 1 Strength."

I like the idea of the attack better. For the power, the "if you killed an enemy this turn" part feels a bit too niche to me. That said, for the attack, all three of those effects on one card is too much with those numbers. Piercing Wail (-6 temp strength) is a great defensive card and 6 Bind is the same effect. 7 damage on top of that already makes a strong uncommon attack where I'd take every copy I saw. Also getting 1 strength for free when drawn pushes it into too much territory. And getting strength when drawn is probably too good in general; at least, if you wanted to keep that, you'd have to make the other effects of the card underwhelming.

From my perspective, aiming to a simple combination of good effects with reasonable numbers is something that would both fit with your idea for this card and produce something that's strong but reasonably balanced. E.g., `Deal 6 (8) damage. Gain 1 Strength. Apply 3 (4) Bind to ALL enemies. Exhaust.` The combination of damage, a bit of scaling, and aoe Bind on one card means that it's always useful, excelling in fights with multiple enemies or enemies with multi-attacks. That's a lot of fights, and the multi-attacks part means this is great in some of the most important fights (every act 3 boss and both act 4 fights have critically important multi-attacks).

To see that this card is definitely going to be good, compare it Decapitate from the Aggression pack, which reads: `Deal 10 (13) damage. Enemy loses 4 (6) Strength this turn. Exhaust.` Compared to Fall Over, Decapitate has higher damage and 1 (2) more temp strength down, but the debuff is only to one enemy and no strength gain. It's a good card and Fall Over should be competitive with it. A bit worse in single-enemy fights, better in multi-enemy fights, and better when you can get more out of the strength (e.g. in longer fights, or when you have multi-hit attacks, or when combined with effects that benefit from dealing more attack damage).

JDodi:
> From my perspective, aiming to a simple combination of good effects with reasonable numbers is something that would both fit with your idea for this card and produce something that's strong but reasonably balanced. E.g., `Deal 6 (8) damage. Gain 1 Strength. Apply 3 (4) Bind to ALL enemies. Exhaust.` The combination of damage, a bit of scaling, and aoe Bind on one card means that it's always useful, excelling in fights with multiple enemies or enemies with multi-attacks. That's a lot of fights, and the multi-attacks part means this is great in some of the most important fights (every act 3 boss and both act 4 fights have critically important multi-attacks).
> 
> To see that this card is definitely going to be good, compare it Decapitate from the Aggression pack, which reads: `Deal 10 (13) damage. Enemy loses 4 (6) Strength this turn. Exhaust.` Compared to Fall Over, Decapitate has higher damage and 1 (2) more temp strength down, but the debuff is only to one enemy and no strength gain. It's a good card and Fall Over should be competitive with it. A bit worse in single-enemy fights, better in multi-enemy fights, and better when you can get more out of the strength (e.g. in longer fights, or when you have multi-hit attacks, or when combined with effects that benefit from dealing more attack damage).

Thanks for the analysis. Alright then, getting 1 Strength at the same time as applying AoE Bind is not negligible, and the damage doesn't need to be very high if there is no on-kill effect. I'll go with that.

src/main/java/thePackmaster/powers/bladestormpack/GaleForcePower.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public void onUseCard(AbstractCard c, UseCardAction action) {
+        if (amount <= 0 || c.type != AbstractCard.CardType.ATTACK) { return; }
+
+        this.flashWithoutSound();
+        addToBot(new VFXAction(new GaleForceEffect(), 0.0F));
+
+        atb(new ApplyPowerAction(player, player, new WindrushPower(player, (c.cost + EXTRA_WINDRUSH) * amount)));
```
modargo, line 42: It's better to use `Wiz.getLogicalCardCost` here and in the check below, so that things like X cost cards and Swivel's free attack power are handled correctly.

src/main/java/thePackmaster/cards/bladestormpack/TradeWinds.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final int COST = 0;
+    private static final int HEAL_IF_COMMON = 2;
+    private static final int GOLD_LOSS_IF_RARE = 5;
+
+    public TradeWinds() {
+        super(ID, COST, CardType.SKILL, CardRarity.UNCOMMON, CardTarget.SELF);
+        baseMagicNumber = magicNumber = HEAL_IF_COMMON;
+        baseSecondMagic = secondMagic = GOLD_LOSS_IF_RARE;
+        exhaust = true;
+
```
modargo, line 31: Since this card can provide healing, it should be tagged as healing so it can't be generated by in-combat card creation effects: `tags.add(CardTags.HEALING);`

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
```diff
@@ -0,0 +1,37 @@
+{
+  "${ModID}:EscalatingBreezePower": {
+    "NAME": "Escalating Breeze",
+    "DESCRIPTIONS": [
+      "Retain up to #b",
+      " card this turn. Chosen Attacks will cost #b",
```
modargo, line 6: I think you updated the wording for the card but not this -- it would be better phrased as `Attacks Retained this way cost #bN less next turn.`

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
```diff
@@ -0,0 +1,37 @@
...
+      " cards this turn. Chosen Attacks will cost #b",
+      " less next turn."
+    ]
+  },
+  "${ModID}:DownwindBlowPower": {
```
modargo, line 11: I don't think this entry is used since Downwind Blow reuses Instant Death's `CloudPower`, so this entry can be deleted.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "${ModID}:GaleForcePower": {
+    "NAME": "Gale-Force",
+    "DESCRIPTIONS": [
+      "Whenever you play an Attack, gain #yWindrush equal to its cost + #b",
+      ", and draw #b",
+      " card if it costs #b",
+      " cards if it costs #b",
+      " or more. NL NL ",
+      "  Windrush : Whenever you would receive unblocked Attack damage this turn, reduce it that much."
```
modargo, line 27: Instead of adding a description of Windrush to the power here, it would be better to add it to the card.

You can do this by overriding `getCustomTooltips` in the card class for `GaleForce`:
```
    @Override
    public List<TooltipInfo> getCustomTooltips() {
        List<TooltipInfo> tooltips = new ArrayList<>();
        tooltips.add(new TooltipInfo(BaseMod.getKeywordProper("anniv5:windrush"), BaseMod.getKeywordDescription("anniv5:windrush")));
        return tooltips;
    }
```

There are examples of this in Packmaster that you can check out, just search for `getCustomTooltips`.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
```diff
@@ -0,0 +1,37 @@
...
+    ]
+  },
+  "${ModID}:WindrushPower": {
+    "NAME": "Windrush",
+    "DESCRIPTIONS": [
+      "Whenever you would receive unblocked Attack damage this turn, reduce it by #b",
```
modargo, line 33: Going by base game powers like Static Discharge I believe this should be `attack` (lowercase) rather than `Attack`. Conceptual difference being that `Attack` is usually reserved for referring to player Attack cards, whereas enemies `attack` the player.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Relicstrings.json
```diff
@@ -0,0 +1,13 @@
+{
+  "${ModID}:FlyingDutchman": {
+    "NAME": "Flying Dutchman",
+    "FLAVOR": "You'd swear it glows and hovers when the weather's real bad.",
+    "DESCRIPTIONS": [
+      "Each turn, after playing #b",
+        " Attack, the next one  grants #b",
```
modargo, line 7: Very minor but there are two spaces after "one" in this line and the line below. Should only be one space.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/UIstrings.json
```diff
@@ -0,0 +1,21 @@
...
+      "Dj0z"
+    ]
+  },
+  "${ModID}:BladeStormModifiers": {
+    "TEXT": [
+      " NL Gain ",
+      " Strength."
+    ]
+  },
+  "${ModID}:BladeStormPackHat": {
```
modargo, line 15: Don't forget to add a hat and fill this text in.

modargo (changes requested):
I've gone through the PR and left comments. Overall the card functionality is in reasonable shape, but the way you're handling card descriptions that change needs a rework (which I've detailed in the comments). There's also a few miscellaneous issues/potential issues I noted and I've tried to call out where descriptions should be changed to match precedent.

I'll plan to do another pass after the comments are addressed, and I may also do a sample run or two. Looking forward to this being ready to go!

src/main/java/thePackmaster/cards/bladestormpack/DownwindBlow.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final int UPG_DAMAGE = 5;
+    private static final int PRECISION = 9;
+    private static final int UPG_PRECISION = 3;
+    private static final int DAMAGE_PROTECTION_PERCENTS = 50;
+
+    public DownwindBlow() {
+        super(ID, COST, CardType.ATTACK, CardRarity.RARE, CardTarget.ENEMY);
+        baseDamage = damage = DAMAGE;
+        baseMagicNumber = magicNumber = PRECISION;
+        baseSecondMagic = secondMagic = DAMAGE_PROTECTION_PERCENTS;
```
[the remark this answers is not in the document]
JDodi: This was in case 50% was deemed too high or too low. I'm fixing that now.

src/main/java/thePackmaster/powers/bladestormpack/EscalatingBreezePower.java
```diff
@@ -0,0 +1,78 @@
...
+import thePackmaster.actions.bladestormpack.EscalatingBreezeAction;
+import thePackmaster.powers.AbstractPackmasterPower;
+
+import java.util.ArrayList;
+
+import static com.megacrit.cardcrawl.dungeons.AbstractDungeon.player;
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+//REFS: RummageDrawAction (packmaster), RunedMithrilRod (warlockpack), RetainCardsAction (base game)
+public class EscalatingBreezePower  extends AbstractPackmasterPower {
```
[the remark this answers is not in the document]
JDodi: I might have, but I'll make sure.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
JDodi: That's the previous wording indeed, forgot that there. Fixing.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
JDodi: I was quite sure that this was possible, just didn't manage to figure it out. Will do.

JDodi:
After multiple runs of testing at A20, I have almost never been happy to find Feeding Winds in my hand. I'm changing the active effect to "Exhaust a card to gain 1 Energy", which has synergy with the Startup effect (via thinning) and is generally more useful than the current active, if that's okay.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "${ModID}:Overpressure": {
+    "NAME": "Overpressure",
+    "DESCRIPTION": "Gain !B! Block. NL Gain !M! Frail. ",
+    "EXTENDED_DESCRIPTION": ["${ModID}:First:", " Gain !B! Block.", " NL "],
+    "FLAVOR": "So... you dare to defy me?!"
+  },
+  "${ModID}:TradeWinds": {
+    "NAME": "Trade Winds",
+    "DESCRIPTION": "Discover an Attack. NL If it is common, heal !M! HP, or if it is rare, lose !${ModID}:m2! gold. NL Exhaust.",
```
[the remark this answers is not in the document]
JDodi: For some reason, it keeps displaying as "anniv5:Discover" in base and upgraded version. I tried adding a space before it, no difference.

src/main/java/thePackmaster/cardmodifiers/bladestormpack/GainStrengthModifier.java
```diff
@@ -0,0 +1,34 @@
...
+    public GainStrengthModifier(int valueIn) {
+        super(valueIn);
+    }
+
+    public String modifyDescription(String rawDescription, AbstractCard card) {
+        return rawDescription + CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("BladeStormModifiers")).TEXT[0] + this.value + CardCrawlGame.languagePack.getUIString(SpireAnniversary5Mod.makeID("BladeStormModifiers")).TEXT[1];
+    }
+
+    public void onUse(AbstractCard card, AbstractCreature target, UseCardAction action) {
+        for(int i = 0; i < this.value; ++i) {
```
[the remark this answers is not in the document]
JDodi: That's clearly better. Done.

src/main/java/thePackmaster/powers/bladestormpack/EscalatingBreezePower.java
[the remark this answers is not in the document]
JDodi: It is broken.
Test with Well-Written Script: first Escalating Breeze happens, then Well-Written Script happens (the cards retained by Escalating Breeze cannot be selected, which is good). The chosen cards are all properly retained, but the cost reduction never happens.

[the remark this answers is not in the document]
JDodi: Okay the cost reduction is always the 2nd set of retained cards. But upgraded Breeze (retains 2) gets only 1 retain (and reduces its cost properly), while the every-turn retain gets the 2 retains (instead of 1) and doesn't reduce cost. That's what led me to believe that the cost reduction was broken.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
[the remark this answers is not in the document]
modargo: You may need to merge in the most recent changes from `erasels:main`, since the keyword is something I added after you started work on this pack.

modargo:
> After multiple runs of testing at A20, I have almost never been happy to find Feeding Winds in my hand. I'm changing the active effect to [0 cost] "Exhaust a card to gain 1 Energy", which has synergy with the Startup effect (via thinning) and is generally more useful than the current active, if that's okay.

That seems fine. Though maybe it's worth thinking up some other effect to give it that does more to work with the rest of the pack? Like upgrading all attacks in your hand, or giving the "Gain 1 Strength" modifier to more cards, or giving ~5 Precision, or something along those lines (those are just random ideas that came to mind, feel free to come up with something else).

modargo:
> > After multiple runs of testing at A20, I have almost never been happy to find Feeding Winds in my hand. I'm changing the active effect to [0 cost] "Exhaust a card to gain 1 Energy", which has synergy with the Startup effect (via thinning) and is generally more useful than the current active, if that's okay.
> 
> That seems fine. Though maybe it's worth thinking up some other effect to give it that does more to work with the rest of the pack? Like upgrading all attacks in your hand, or giving the "Gain 1 Strength" modifier to more cards, or giving ~5 Precision, or something along those lines (those are just random ideas that came to mind, feel free to come up with something else).

Or if you want to stick with the "exhaust a card to gain energy" idea, it could gain more energy if you exhaust an attack or status (I threw in status there because it helps in all the fights where enemies give you statuses and provides some interesting cross-pack synergy). Basically, spice it up just a bit!

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
JDodi: I still find it highly confusing. Adding this code to the card gives it a 2nd Windrush tooltip, while adding it to the power gives an error immediately. Looking at `getCustomTooltips` examples, I realize I had already looked through them, and it's still opaque to me. Gulp (ingame name Meal Prep) gives Ablation, Slaytabase doens't know what that is so i went and had a look but the Ablation buff (technically power) only has a tooltip for Ablation, not for Temporary HP too, which would be what I need to follow to add a Windrush tooltip to GaleForcePower, right? I'm trying to find other examples now.

JDodi:
> > That seems fine. Though maybe it's worth thinking up some other effect to give it that does more to work with the rest of the pack? Like upgrading all attacks in your hand, or giving the "Gain 1 Strength" modifier to more cards, or giving ~5 Precision, or something along those lines (those are just random ideas that came to mind, feel free to come up with something else).
> 
> Or if you want to stick with the "exhaust a card to gain energy" idea, it could gain more energy if you exhaust an attack or status (I threw in status there because it helps in all the fights where enemies give you statuses and provides some interesting cross-pack synergy). Basically, spice it up just a bit!

I'd love to, but with the Startup effect taking up 3 lines straight away, the text real estate on the card is very limited before hitting the 6 lines limit so any sort of condition or nuance is hard to fit in.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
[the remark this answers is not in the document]
JDodi: Just checked, and my branch is up to date and 34 commits ahead, so I must have the fix already. Trying out various things to fix it.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/UIstrings.json
JDodi: It appears that I'll have to draw it first. Art not being my forte, this is going to take some time

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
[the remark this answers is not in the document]
JDodi: In fact the other cards using Discover right now (Mimicry, Paintbrush) have no tooltip for it and it's not yellow in their text. I should have checked that first. I'll leave it be for now.

JDodi:
Here's an idea: since what I'd want it to do ideally just resembles what Gale-Force does... why not just cut to the chase, raise the cost to 1 E, and make the tradeoff of the exhaust a Gale-Force card (always unupgraded) in hand? it makes the "pop off to victory" part of the fight more reliable to reach, while gating it behind paying 3 E for a 2E(1E upgraded) power. It can also be used to repeatedly exhaust the Gale-Force it has printed itself, if there's any exhaust (or card creation) synergy going on.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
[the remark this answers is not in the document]
modargo: That indicates you don't have the most recent changes where I added the keyword.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
modargo: I see why, I didn't notice Gale-Force already had Windrush as a keyword on it. Given that, you can ignore what I said about `getCustomTooltips`, and just delete the Windrush explanation here. Someone who is playing with Gale-Force in their deck has the opportunity to read the card before they play it to understand Windrush, and to look at their master deck during combat to double check. So it's fine for the power to not explain it again.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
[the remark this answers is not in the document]
modargo: For what it's worth, I've pulled and built the code for this PR and it looks fine to me. So not sure what's going on for you, maybe you have something cached and need to do a clean build? Regardless, it will almost certainly work when this gets merged in, so you don't need to worry about it.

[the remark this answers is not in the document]
JDodi: Alright, I better find out why I'm not getting the latest changes and fix that though.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Powerstrings.json
JDodi: Roger that, I didn't think about these possibilities, that simplifies things.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
```diff
@@ -0,0 +1,57 @@
...
+    "EXTENDED_DESCRIPTION": ["Discover an upgraded Attack. NL If it is common, heal ", " HP, or if it is rare, lose ", " gold. NL Exhaust."],
+    "FLAVOR": "Sword! Return to your master!"
+  },
+  "${ModID}:TempestOfStrikes": {
+    "NAME": "Tempest of Strikes",
+    "DESCRIPTION": "Deal !D! damage and apply !M! ${ModID}:Bind a number of times equal to half of the number of Attacks in your deck (",
```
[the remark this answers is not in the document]
JDodi: Hmm I don't see a way to do exactly that, because of the Exhaust which should by convention be the last thing in text. But I think I can at least make the "0 times" disappear outside of combat, so I'm doing that.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Cardstrings.json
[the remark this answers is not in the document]
modargo: Have you pulled and built the main repo? If you haven't, that's probably why.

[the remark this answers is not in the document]
JDodi: Packmaster Character Main? I thought I was only supposed to touch Expansion Packs. Is that where the Discover keyword is? I'm confused. Well I just went over there and sure enough it mentions Discover. Trying to figure out what to do here

JDodi:
It still says "Changes Requested" but I believe everything has been addressed (excluding stretch goals).

After another 2 runs (both dead in act 3 a20 to elite/boss) I am quite happy with the state of almost every card right now.
The pack has clear weaknesses like taking strong singular hits, which is fine. The only thing bothering me is that even with significant investment into this pack, I still find the uptime on Windrush lacking when Gale-Force couldn't be obtained.

I don't want to make many more cards give the buff since that's a design pitfall, but maybe `Whenever you attack, Gain 1 Windrush` could be baked into the buff itself?
That way, once you're buffed, you may play 1 to 3 Attacks per turn to maintain between 2 and 6 stacks in the long run . Math says it would be very difficult to maintain any more than that without extra sources due to the rounded up decay. I don't think any extra improvement or card source would be needed, in fact existing sources could be reduced by 25-50%. Also makes synergy with other packs better since you wouldn't need heavy investment into the pack.
Thoughts on this?

modargo:
Yeah, I'll take a second pass through the PR in the next day or two and see if there's anything else that should be changed.

For Windrush, my first reaction is that it seems okay for it to be hard to keep up the buff without Gale-Force. There are a variety of pack mechanics where having constant uptime is difficult to achieve, and it's okay for Windrush to be like that.

I don't really like the idea of having attacks add stacks since most decks play attacks on most turns, so that would be pretty easy to keep up. Before anything like that, I'd want to consider adding Windrush to more cards and/or slightly increasing the amounts. One or two more cards with it would make an impact (and hopefully wouldn't over-saturate the pack). The obvious candidates to me are the two commons; either Flying Swords or Overpressure could give 1 or 2 Windrush with only minor tweaks to the other numbers on the card.

JDodi:
Alright, I'll start looking at where to add/increase the sources. Certainly on Overpressure. Probably not on using Flying Swords since that would be a lot of effects for a common, and I really like how the Startup effect reliably gives that window of relative safety to setup early on. So I'll think of another source too.
Should I push anything I come up with, or wait until you've given the pack some field testing?

modargo:
> Alright, I'll start looking at where to add/increase the sources. Certainly on Overpressure. Probably not on using Flying Swords since that would be a lot of effects for a common, and I really like how the Startup effect reliably gives that window of relative safety to setup early on. So I'll think of another source too. Should I push anything I come up with, or wait until you've given the pack some field testing?

I agree that Overpressure seems like a better fit than Flying Swords between the two commons.

Go ahead and push any changes you make, iterating during playtesting is normal. I've done one run with the pack but unfortunately saw very few cards from it so I wasn't able to form any meaningful conclusions. When I do, having whatever changes you want to make is better than not having them.

modargo:
Just did a playtesting run with this pack. The other packs were For The Content, Jeweler's Kit, Monster Hunter, The Swarm, and Utility.

I took a Fall Over early and it was great. Flying Swords was also pretty good, decent damage and the Startup Windrush saved me some HP over the run. I saw the new version of Overpressure a few times in act 1 card rewards but didn't take it, the lack of block (outside of the First effect) made me think it wasn't worth it.

Downwind Blow was the first rare I saw, about halfway through act 1, and I took it to try it out. It immediately did very strong things against Sentries -- it effectively blocked for 5 on the turn I played it, then I used the Precision with a Smithing Hammer. I also took a Tempest of Strikes in act 2, which was another great way to take advantage of the Precision from Downwind Blow. It was also just very strong in general, it was effectively a full block in many fights (Bind ~12 handles almost anything until you reach act 3).

I also got Pen Nib from the first elite, which made all the previously mentioned cards even better, since I could line up super high damage turns. Definitely a great start for this pack. It would have been nice to get Snecko Eye too after those picks, but instead I had to make do with using Lapis Lazuli (from the gems pack) to make Downwind Blow Occult, so I could always play it at the cost of it exhausting.

Highlights of the rest of the run:
* I put a few of the Frost gems on attacks, so I had multiple ways to accumulate block and other defenses while dealing damage
* I got two copies of Rhapsody (from the Content Creators pack), so the multiplicative damage scaling got a bit crazy
* I took Cursed Key and Sozu as my boss relics, so I had tons of energy, and took an Unrelenting Form as way to scale
* I ended up with enough attack synergy that hallway fights got destroyed very quickly, and had enough defense that I had no problem surviving into my scaling kicked in for elites and bosses
* I never focused on keeping Windrush up, I just treated it as a nice to have when it was convenient

The Heart fight was close but I won. Everything prior to the Heart was very easy since I could kill enemies so quickly, but I didn't draw the right cards to full block the Heart's first attack and that took me to low HP. But I was able to neutralize the multi-attack and stabilized from there. Didn't have any trouble with the rest of the fight.

My perspective on the pack's cards so far:
* Flying Swords is a fine common attack
* Fall Over is a very strong uncommon attack. I think it's fine but it's one of the cards where taking every copy you see will often be the correct play (which I thought might be the case; playing a run with it confirmed that). If I were balancing this pack I'd probably reduce it to 2 Bind.
* Tempest of Strikes is very strong and I think it needs a nerf. The values get too high too easily, and it's able to either kill or neutralize a ton of enemies. It's like if Dash typically did 16+ damage, had a Strength scaling multiplier of 4+, and completely neutralized multi-attacks from the enemy hit by it. Sure, it exhausts, but the turn you play it is almost always safe. And it requires having some attacks in your deck, but that's not a big deal with a pack that encourages taking some attacks (even when you're removing Strikes, by the time you do that you normally have other attacks). There's a variety of ways to adjust this, but really should be toned down.
* I'd be more inclined to take Overpressure if it still had block for its base effect. Maybe it could give a mix of block and Windrush?
* Downwind Blow was quite good through the whole run. The mix of effeects meant it always had something to contribute, especially with multi-attacks. Pictured: playing it with 2 Strength, Aggression stance, Pen Nib, Inspiration 150, and the enemy Vulnerable.
* Other cards: I saw Trade Winds, but other cards in the rewards seemed better. I also saw Escalating Breeze, but I already had 5 energy so it wasn't an attractive pick. Never saw Geo Da Ray or Gale-Force.

I enjoyed playing with the pack, so feels like things are good on that front and it just needs some tweaks and adjustments based on playtesting.

src/main/java/thePackmaster/cards/bladestormpack/FeedingWinds.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public void upp() {
+        upgradeMagicNumber(UPG_STRENGTH_CARDS);
+        upgradeSecondMagic(UPG_WINDRUSH);
+    }
+
+    //From here down, dynamic Startup tooltip based on AbstractPickThemAllCard (pickthemallpack).
+
+    public void onCreateCard(AbstractCard c) {
```
modargo, line 81: A small thing, but as a method that you've implementing from an interface, it's best practice to mark this with `@Override` (just like all the other methods that are originally defined somewhere other than this class). Same comment applies to Flying Swords.

JDodi:
That's very informative! 
I'll take another look at the numbers and see what I can do about the points you brought up.

src/main/java/thePackmaster/cards/bladestormpack/AbstractBladeStormCard.java
```diff
@@ -0,0 +1,24 @@
...
+public abstract class AbstractBladeStormCard extends AbstractPackmasterCard
+{
+    public AbstractBladeStormCard(String cardID, int cost, CardType type, CardRarity rarity, CardTarget target)
+    {
+        super(cardID, cost, type, rarity, target, "bladestorm");
+
+    }
+
+    @Override
+    public void upp() {
```
modargo, line 16: Having this is redundant and I'd recommend deleting it. Trade Winds is the one card that doesn't do anything for `upp`, and it overrides `upp` itself (which is the better approach, since normally cards will need to override this).

Since you've already coded all the cards it doesn't make a significant difference, but deleting this is better in case you (or someone else) copies this as the basis for a new pack, or if you decide to deprecate one of the cards in this pack and make a new pack. Plus the general benefit of having Packmaster code consistently follow best practices and avoid redundant code.

src/main/java/thePackmaster/powers/bladestormpack/GaleForceDrawPower.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public void onUseCard(AbstractCard c, UseCardAction action) {
+        if (amount <= 0 || c.type != AbstractCard.CardType.ATTACK || Wiz.getLogicalCardCost(c) < DRAW_COST_THRESHOLD) {
+            return;
+        }
+
+        this.flashWithoutSound();
+        addToBot(new VFXAction(new GaleForceEffect(Color.LIGHT_GRAY, true), 0.0F));
+
+        if (Wiz.getLogicalCardCost(c) >= DRAW_COST_THRESHOLD) {
```
modargo, line 39: Given the if statement at the start of this function, any card that reaches here already must have cost >= the threshold, so you can get rid of this if statement.

src/main/java/thePackmaster/powers/bladestormpack/WindrushPower.java
```diff
@@ -0,0 +1,52 @@
...
+        return damageAmount;
+    }
+
+    @Override
+    public void atEndOfRound() {
+        //Cut in half, rounding down ( = rounding the reduction up) and able to fully fade out.
+        if (amount == 1) {
+            addToBot(new RemoveSpecificPowerAction(this.owner, this.owner, this));
+        } else if ((amount % 2) != 0){
+            addToBot(new ReducePowerAction(this.owner, this.owner, this, (amount+1) / 2));
```
modargo, line 43: You can simplify this function down to only this single line. With integers, `(amount + 1) / 2` produces the correct result for both even and odd numbers, and `ReducePowerAction` handles removing the power if the reduction is greater than or equal to the current amount of the power. So this one case handles it all and the other cases are unnecessary.

src/main/resources/anniv5Resources/localization/eng/bladestormpack/Relicstrings.json
```diff
@@ -0,0 +1,13 @@
+{
+  "${ModID}:FlyingDutchman": {
+    "NAME": "Flying Dutchman",
+    "FLAVOR": "You'd swear it glows and hovers when the weather's real bad.",
+    "DESCRIPTIONS": [
```
modargo, line 5: Looking at the wording for this relic, it would be a bit easier to parse if the wording was closer to Akabeko or Necronomicon. E.g. `The second time you play an Attack each turn, gain 2 Windrush and 3 Precision.`

Also, think you can get rid of the pluralization logic, since the number of attacks to trigger this is fixed.

src/main/java/thePackmaster/relics/bladestormpack/FlyingDutchman.java
```diff
@@ -0,0 +1,85 @@
...
+
+        if (counter >= ATTACKS_FOR_TRIGGER) {
+            Wiz.applyToSelf(new WindrushPower(player, WINDRUSH));
+            Wiz.applyToSelf(new Precision(player, PRECISION));
+
+            pulse = false;
+            addToTop(new RelicAboveCreatureAction( player, this));
+            flash();
+
+            this.grayscale = true;
```
modargo, line 57: I'm pretty sure the base game only uses `grayscale` for relics that no longer do anything for the rest of combat. Think Fossilized Helix or Horn Cleat.

Instead of using grayscale for this, you can take the same approach as Necronomicon and have a `private boolean activated;` field. Basically just make that field and switch out all the uses of grayscale to use that field instead, and everything should work fine.

JDodi:
- Glad you like Flying Swords, I also feel it's in a good place.
- I'll reduce Fall Over's Bind by 1, I don't know what magic you do with it but it doesn't feel as strong to me. Maybe some copy shenanigans?
- Tempest of Strikes: I already nerfed it once after the "every 2 attacks" change, but it is strong, so here we go. However, I suspect that stacking up on random, unupgraded attacks to fully juice it may have been a reason why some of my runs suddenly got cut short by Nemesis and company, risk-reward basically.
- About Overpressure, I'm mostly worried that it would look too complex as Gk would say, if the non-First effect also has block (that would be 3 things in the normal effect, then there's the First) so I guess I'll raise the Windrush until it looks competitive with other Commons. The block portion is already quite high, though it tends to be easily nullified by Autoplay and similar effects.
- Downwind Blow: not really about the card which seems quite fine, but I did realize that the pack has no Weak or Vulnerable at all, which is another of its weaknesses (especially an attack pack)
- Trade Winds: if the other cards in said rewards were also Uncommons, that would mean it needs better numbers to compete? Let me know if that wasn't the case. I'm thinking "heal 6 HP" (since after all, it's not guaranteed, but higher sounds OP) and/or maybe 3-4 gold for rares?

I'll go ahead and push this ASAP, keeping my ears open to further feedback.

And BTW, props on the Heart kill, I still haven't managed. Even had Pen Nib, multiple Frost orb sources and generally a good setup in most of the runs I'd say. Just couldn't mitigate big hits when needed a couple times. I'll try with your exact setup in the run described above to get some insight.

modargo (commented):
Took a second pass through the code. Noted a few code and wording changes to make, all relatively minor. Once those are addressed this should be good on that front. Then it's just a question of how much more playtesting and iteration you end up wanting to do.

src/main/java/thePackmaster/powers/bladestormpack/GaleForceDrawPower.java
JDodi: Right, at some point I forgot I had added it to the top

modargo:
> * I'll reduce Fall Over's Bind by 1, I don't know what magic you do with it but it doesn't feel as strong to me. Maybe some copy shenanigans?

Aggression helped (Utility pack has a good card for entering it, and I got another Aggression card from the PM Booster Pack relic). That and two copies of Smithing Hammer. Multi-attacks good. It was also just a strong run.

> * Tempest of Strikes: I already nerfed it once after the "every 2 attacks" change, but it is strong, so here we go. However, I suspect that stacking up on random, unupgraded attacks to fully juice it may have been a reason why some of my runs suddenly got cut short by Nemesis and company, risk-reward basically.

Yeah, I don't think it's worth taking mediocre attacks just to juice it up. I'm sure there will be some people who don't get much value from it because they tend to build small decks or dislike taking attacks at all. In my case, I happily build bigger decks (they're more resilient to statuses and Packmaster runs usually have the draw available to support finding the cards you need) and I take advantage of all the Packmaster attacks that have effects that aren't just damage (even a common like Smithing Hammer can be worth taking in act 2 or 3 if the upgrades will consistently get value). Fun statistic: my average deck size for a winning Packmaster run is 34 cards.

> * Trade Winds: if the other cards in said rewards were also Uncommons, that would mean it needs better numbers to compete? Let me know if that wasn't the case. I'm thinking "heal 6 HP" (since after all, it's not guaranteed, but higher sounds OP) and/or maybe 3-4 gold for rares?

IMO Trade Winds is fine, just was up against Scarab Shield+ and that was the clear pick. Big dense block cards that come already upgraded are hard to beat.

modargo:
Oh right, meant to post this:
image

JDodi:
That's one proper juiced attack!

modargo:
Everything looks good so I'll go ahead and merge this. Congratulations on contributing your second pack!

PR #96, opened 2025-03-19
Balance doppel pack: nerf some cards

herbix:
Concatenation: Remove "Draw 1 card".
Resistance: Remove "Draw 1 card".
Summon: Reduce card draw: 2(3) -> 1(2).
Tides: Remove "Summon 1 Doppel". Damage value change: 6(8) times 2+N -> 5(7)+N times 3.
Vortex: Reduce damage: 9(12) -> 8(11).
Worm Hole: Doesn't stack.


erasels:
Thanks for the balance adjustments and updating the ZHS strings as well!

PR #106, opened 2025-07-06
Siege pack partial rework

JDodi:
This rework changes a few cards and brings minor improvements.
- Major changes to 2 of the least satisfying cards. Shelling's code is much simpler.
- Made Shells generally better at the cost of slower ramp-up to follow the pack's theme. Slightly changed Shell Forge to fit this.
- Lowered Shrapnel buff icon flashing spam.


JDodi:
SiegePackCompendium_v9 5
SiegePackCompendium(U)_v9 5
Proposed rework compendiums

src/main/java/thePackmaster/cards/siegepack/ThinkTwice.java
```diff
@@ -1,25 +1,27 @@
...
 import static thePackmaster.cards.siegepack.FlavorConstants.*;
 
 public class ThinkTwice extends AbstractSiegeCard {
     public final static String ID = makeID("ThinkTwice");
-    private static final int COST = 2;
-    private static final int BLOCK = 6;
-    private static final int UPGRADE_BLOCK = 4;
-    private static final int BLOCK_PER_ATTACKER = 5;
+    private static final int COST = 1;
+    private static final int BLOCK = 16;
```
erasels, line 15: I think we need to reduce this a bit if the shell gain is there. Can't make it go nicely then, but I think having it be 14, -7 and then -4 or something

src/main/resources/anniv5Resources/localization/eng/siegepack/Keywordstrings.json
```diff
@@ -12,6 +12,6 @@
       "shell",
       "shells"
     ],
-    "DESCRIPTION": "Attacks deal #b5 additional damage to enemies and consume #b1 #yShell."
+    "DESCRIPTION": "Attacks deal #b3 additional damage to enemies and consume #b1 #yShell. NL #yShells get stronger at the start of each turn after you first gained any."
```
erasels, line 15: This seems like a pain to keep track of. I don't think this is a good idea

erasels (changes requested):

src/main/resources/anniv5Resources/localization/eng/siegepack/Keywordstrings.json
JDodi: Oh, alright. This was sort of a central point of the rework, to make Shells more interesting. At that point I think I'll draft up something that's Vigor-based. The extra synergy with the 13 Packmaster Vigor cards i could find should make it more interesting than reverting to current Shell design. What do you say?

src/main/java/thePackmaster/cards/siegepack/ThinkTwice.java
JDodi: This was the closest nice block number i could find south of 18 (the number before it had shell gain), but I understand. 
What if it costs 2 like pre-rework, with 28 block -14 on use? Upgraded would be -7. Alongside the Shell->Vigor rework i mentioned in the feedback just below this, the Vigor amount could be tuned more finely than Shells for balance. 
Since it's almost a curse without exhaust when depleted, i think it would match similar PM cards like Fear, Freedom or Showoff.

src/main/java/thePackmaster/cards/siegepack/ThinkTwice.java
```diff
@@ -28,26 +30,20 @@ public ThinkTwice() {
...
-    public void triggerOnGlowCheck() {
-        this.glowColor = AbstractCard.BLUE_BORDER_GLOW_COLOR.cpy();
-        for (AbstractMonster m : Wiz.getEnemies()) {
-            if (Wiz.isAttacking(m)) {
-                this.glowColor = AbstractCard.GOLD_BORDER_GLOW_COLOR.cpy();
-                break;
-            }
-        }
+    //Base game's ModifyBlockAction seems bugged at 0 block.
+    public void reduceBlock () {
```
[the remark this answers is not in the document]
JDodi: Alright, and yes

src/main/resources/anniv5Resources/localization/eng/siegepack/Keywordstrings.json
JDodi: The non-scaling splash damage shells were indeed worse by the way

src/main/java/thePackmaster/cards/siegepack/ThinkTwice.java
erasels: I mean, compare those values with Imprevious on ironclad, a very good Block card. You argue that it becomes a cruse after 2 plays, which is true but other than that this is basically the equivalent of a rare ironclad block card on firwst place and then a still good block card with secondary effect on second.

src/main/resources/anniv5Resources/localization/eng/siegepack/Keywordstrings.json
erasels: I think going for Vigor could be cool, there's still unexplored space there, although that might be hard to fit flavor wise.

If you want to keep shells and make them more interesting, the best I can think of to stay close to your proposed change is making Shells add 3 + stack amount additional damage and be consumed per hit. 
Example:
You have 7 shells and play Pummel (4 hits of 2 damage)
hit 1: 2 + 7 (7 shells)
hit 2: 2 + 6 (6 shells)
... etc

JDodi:
The Vigor rework should now be complete. I'll wait for feedback before making other changes.

JDodi:
No worries, I keep thinking about how much work maintaining PM is, and that's just 1 of the mods you're involved in. Also, it was summer, and so on.

I'm all for simplification!

Logistics: that will be done.

Think Twice: sure, why not. If the amount of vigor should go down because of that, let me know.

Ballistic Strike: I'll change the second line like you said. For the rest I'd rather not change it, countering attacks is half the fun of the card, and it's the player's job to find tools to clear squishy targets and retain/fetch the card to play it at the best possible time for maximum benefit. For in-pack synergy and the "few but big" attacks theme the 10 base damage is also pretty important.

Shell Forge: I kept going back and forth between versions and that seemed the most interesting, but what you said was also nice to play with and had easily math-able output. I'll go back to that with the numbers it had when i tested it. This one was fun while it lasted.

erasels:
Alright, those changes (fair enough on Ballistic Strike) should be it then

JDodi:
I believe everything works and looks fine now

src/main/java/thePackmaster/powers/siegepack/RampingTurnVigorGainPower.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private final int increaseFix;
+
+    public RampingTurnVigorGainPower(AbstractCreature owner, int amount, int increase) {
+        super(POWER_ID, NAME, PowerType.BUFF, true, owner, amount);
+        this.amount = amount;
+        this.amount2 = increase;
+        this.increaseFix = increase;
+
+        this.priority = 8000; // EventHorizonPower priority should be 5 by default.
```
erasels, line 31: Is highest priority last?

erasels (changes requested):

erasels:
Looks good, it's done!

PR #118, opened 2025-08-23
Tf2 pack

Quorrian:
tf2 themed pack about adding persist to attacks and building flanking to gain block on hit


Quorrian:
> The merge looks fine to me.
> 
> Please also add a screenshot of the cards in the compendium for easier reference

image

Quorrian:
something broke :|  I'll have to go fix it

Quorrian:
I think it works like it should now

erasels:
Alright, after talking about it with Modargo a bit, I managed to narrow down why I think this pack needs a few changes to really come together.

The problem is twofold, you singularly focus on Attacks and Block which means there's no real support for the thing your main mechanic is trying to achieve (no real draw or energy generation, no debuffs, etc). The thing that compounds this is the distribution of effects, you've got 2 ammo cards in uncommon and 4 attacks spread between the first two rarities which basically all ask you to use the mechanic (that is not supported much outside of this pack) to get maximum value out of them.

A good tweak modargo suggested was removing dispenser (powers that generate a token per turn are generally just too slow) and making Ammo a common 1 cost card (or a card like Cloak and Dagger). That would give easier access to the most important mechanic this pack offers. Big Earner could also become an ammo card of some kind so there's a good spread throughout the rarities. Ammo box would obviously need a small change if Ammos cost energy.

2 3 cost powers in rares is also not a great idea as this pack is already energy starved as is, combine that with Cozy Camper being a bit of a weird card from the get go (pretty hard to control) and that might be another good target to change up.

It'd also be nice if one of the first 4 attacks could be a little more support focus instead of being a straight damage scaler.

Quorrian:
Reworked mini sentry to be 0 cost 2 damage 2(3) times. On resupply apply 1 vuln to all enemies

Quorrian:
when testing i noticed there wasn't a thumbnail for the pack. I don't really know where that goes

PR #121, opened 2025-08-29
Runic Artifacts Pack

Dayvig:
Combining orbs and stances into one wonderful package!!!


src/main/java/thePackmaster/packs/RunicPack.java
```diff
@@ -0,0 +1,40 @@
...
+    private static final UIStrings UI_STRINGS = CardCrawlGame.languagePack.getUIString(ID);
+    public static final String NAME = UI_STRINGS.TEXT[0];
+    public static final String DESC = UI_STRINGS.TEXT[1];
+    public static final String AUTHOR = UI_STRINGS.TEXT[2];
+    //public static final String CREDITS = UI_STRINGS.TEXT[3];
+
+    public static boolean channeledOrbThisTurn = false;
+
+    public RunicPack() {
+        super(ID, NAME, DESC, AUTHOR, new PackSummary(4, 1, 3, 2, 4, PackSummary.Tags.Stances, PackSummary.Tags.Orbs));
```
modargo, line 21: I'm guessing the ratings here were copied over from another pack.

Here's my impression on ratings from my handful of playtest games:
* Offense: 3. Glyph orbs do a decent job and the pack has several good attacks.
* Defense: 4. Glyph orbs are powerful and the pack has a few good block cards too.
* Support: 3. Not entirely sure about this one. The pack has some draw and also has card generation with Runic stance, and the stances help out a wide range of other strategies (the artifact from Ancient stance offers very flexible benefits).
* Frontload: 3. The pack has decent options, but nothing too special. Though this is probably the pack's weakest aspect, so this could be a 2 instead.
* Scaling: 4. Switching between this packs stances gives scaling (strength/focus) or potential scaling (artifact, which lets you do a lot of powerful stuff if you get temp strength/dex/focus).

src/main/java/thePackmaster/powers/runicpack/ObeliskPower.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.ArrayList;
+
+public class ObeliskPower extends AbstractPackmasterPower {
+
+    public static final String POWER_ID = SpireAnniversary5Mod.makeID("ObeliskPower");
+    private static final PowerStrings powerStrings = CardCrawlGame.languagePack.getPowerStrings(POWER_ID);
+    public static final String NAME = powerStrings.NAME;
+    public static final String[] DESCRIPTIONS = powerStrings.DESCRIPTIONS;
+    ArrayList<AbstractCard> journalLog = new ArrayList<>();
```
modargo, line 24: This isn't being used so you can delete it.

src/main/java/thePackmaster/stances/runicpack/RunicAura.java
```diff
@@ -0,0 +1,15 @@
...
+package thePackmaster.stances.runicpack;
+
+import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.math.MathUtils;
+import com.megacrit.cardcrawl.vfx.stance.StanceAuraEffect;
+
+public class RunicAura extends StanceAuraEffect {
+
+    public RunicAura() {
+        super("Wrath");
```
modargo, line 11: Better to use `WrathStance.STANCE_ID`.

src/main/java/thePackmaster/stances/runicpack/RunicStance.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public RunicStance() {
+        this.ID = STANCE_ID;
+        this.name = CardCrawlGame.languagePack.getStanceString(STANCE_ID).NAME;
+        this.updateDescription();
+    }
+
+    @Override
+    public void atStartOfTurn(){
+        AbstractCard newChannelCard = ExpansionPacks.channelCards.get(AbstractDungeon.cardRng.random(0, ExpansionPacks.channelCards.size()-1)).makeStatEquivalentCopy();
```
modargo, line 56: Is there a reason for `makeStatEquivalentCopy` here instead of just `makeCopy`? It seems to me that `makeCopy` should work fine, and when that's the case I think using `makeCopy` is better (because `makeStatEquivalentCopy` implies "there might be something special about the specific instance of the card").

src/main/resources/anniv5Resources/localization/eng/runicpack/Powerstrings.json
```diff
@@ -0,0 +1,15 @@
...
+    "NAME": "Power of the Obelisk",
+    "DESCRIPTIONS": [
+      "When you change #yStances, Channel a random #yOrb.",
+      "When you change #yStances, Channel #b",
+      " random #yOrbs."
+    ]
+  },"${ModID}:AttackFocusLoss": {
+  "NAME": "Runic Concentration",
+  "DESCRIPTIONS": [
+    "When you play an Attack, lose #b1 #yFocus and reduce by 1."
```
modargo, line 12: I believe Attack should be highlighted here to be consistent with base game powers like Rage.
```suggestion
    "When you play an #yAttack, lose #b1 #yFocus and reduce by 1."
```

src/main/java/thePackmaster/orbs/runicpack/Glyph.java
```diff
@@ -0,0 +1,169 @@
...
+public class Glyph extends CustomOrb {
+    public static final String ORB_ID = SpireAnniversary5Mod.makeID(Glyph.class.getSimpleName());
+    private static final OrbStrings orbString = CardCrawlGame.languagePack.getOrbString(ORB_ID);
+    public static final String NAME = orbString.NAME;
+    public static final String[] DESCRIPTIONS = orbString.DESCRIPTION;
+    private static final String IMG_PATH = makePath("/images/orbs/spherespack/Blaze.png");
+    private static final float SPIRIT_WIDTH = 96.0f;
+
+    private final static int BASE_PASSIVE = 1;
+    private final static int BASE_EVOKE = 4;
```
modargo, line 40: Given that Glyphs effectively get double scaling (extra damage and extra block), what do you think about making the base evoke 3 instead of 4? So instead of getting +3 to one stat like Frost and Lightning, Glyphs get +2 to 2 stats. That still seems fair and Glyph orbs are good enough that I really don't think it would make them too weak. It would also reduce the difference between the passive effect and the evoke effect -- right now at 0 focus the evoke is 4 times the passive!

src/main/resources/anniv5Resources/localization/eng/runicpack/Keywordstrings.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "PROPER_NAME": "Runic",
+    "NAMES": ["runic"],
+    "DESCRIPTION": "At the start of the turn add a card which #yChannels orbs to your hand with #yEthereal and #yExhaust. When exited, gain #b1 #yFocus."
```
modargo, line 5: Should be a comma after "at the start of your turn". I'd also suggest "that" instead of "which" (both here and in the stance strings).
```suggestion
    "DESCRIPTION": "At the start of the turn, add a card that #yChannels orbs to your hand with #yEthereal and #yExhaust. When exited, gain #b1 #yFocus."
```

src/main/resources/anniv5Resources/localization/eng/runicpack/Powerstrings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "${ModID}:ObeliskPower": {
+    "NAME": "Power of the Obelisk",
+    "DESCRIPTIONS": [
+      "When you change #yStances, Channel a random #yOrb.",
```
modargo, line 5: Base game powers (and cards) that channel orbs use "1" instead of "a" for the singular case (see e.g. Static Discharge, Storm).
```suggestion
      "When you change #yStances, Channel 1 random #yOrb.",
```

Dayvig:
Updated PR!

erasels (changes requested):
I think Runic does a lot of the lifting for this pack to be an Orb and Stance pack. Generally, it looks alright? I see how most cards are generally useful instead of focusing too strong on either archetype so it prevents the associated issues.

src/main/resources/anniv5Resources/localization/eng/runicpack/Keywordstrings.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "PROPER_NAME": "Runic",
+    "NAMES": ["runic"],
+    "DESCRIPTION": "At the start of the turn, add a card that #yChannels orbs to your hand with #yEthereal and #yExhaust. When exited, gain #b1 #yFocus."
+  },
+  {
+    "PROPER_NAME": "Glyph",
+    "NAMES": ["glyph", "glyphs"],
+    "DESCRIPTION": "Orb: At the start of your turn, gain #b1 Block and deal #b1 damage to a random enemy. When #yEvoked, gain #b4 Block and deal #b4 damage to a random enemy."
```
modargo, line 10: Need to update the keyword for the new values. I'll do it with Github's web interface.
```suggestion
    "DESCRIPTION": "Orb: At the start of your turn, gain #b1 Block and deal #b1 damage to a random enemy. When #yEvoked, gain #b3 Block and deal #b3 damage to a random enemy."
```

PR #147, opened 2026-03-18
Adding Rick and Morty expansion pack

grantw99:


src/main/java/thePackmaster/cards/rickandmortypack/PortalGun.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+        Wiz.doDmg(abstractMonster, this.damage, AbstractGameAction.AttackEffect.POISON);
+        if (abstractPlayer.hasPower(PortalSicknessPower.ID)) {
+            Predicate<AbstractCard> zeroCostAttackPredicate = card -> 0 == card.cost && CardType.ATTACK.equals(card.type);
+            AbstractCard randomZeroCostAttack = Wiz.getRandomItem(Wiz.getCardsMatchingPredicate(zeroCostAttackPredicate));
+            if (randomZeroCostAttack == null) {
+                // This means that there were no zero cost attacks in the card pool. Will give random 0 cost if that's the case
+                randomZeroCostAttack = Wiz.getRandomItem(Wiz.getCardsMatchingPredicate(zeroCostAttackPredicate, true));
```
grantw99, line 31: I feel like this is the best way to handle this situation, but would like input if that's not the consensus. I don't feel like the player should be penalized for the card pool they've chosen not having any zero cost attacks, and the card doesn't specify that it will always be from the current card pool, but I feel like it's better to try to give from the pool when we can

src/main/java/thePackmaster/packs/RickAndMortyPack.java
```diff
@@ -0,0 +1,37 @@
...
+
+public class RickAndMortyPack extends AbstractCardPack {
+    public static final String ID = SpireAnniversary5Mod.makeID(RickAndMortyPack.class.getSimpleName());
+    private static final UIStrings UI_STRINGS = CardCrawlGame.languagePack.getUIString(ID);
+    public static final String NAME = UI_STRINGS.TEXT[0];
+    public static final String DESC = UI_STRINGS.TEXT[1];
+    public static final String AUTHOR = UI_STRINGS.TEXT[2];
+
+    public RickAndMortyPack() {
+        super(ID, NAME, DESC, AUTHOR, new PackSummary(3, 2, 4, 4, 3));
```
grantw99: I have no idea how to score these numbers lol

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
+{
+  "${ModID}:KeepSummerSafe": {
+    "NAME": "Keep Summer Safe",
+    "DESCRIPTION": "Gain !B! Block. NL If you have a debuff, draw 2 cards."
+  },
+  "${ModID}:MortysMindBlowers": {
+    "NAME": "Morty's Mind Blowers",
+    "DESCRIPTION": "Remove all debuffs. NL Draw !M! card, it costs 0 this turn. NL Gain !${ModID}:m2! ${ModID}:Portal_Sickness. NL Exhaust."
```
modargo: Small wording change to match cards like Hierophant (from the Arcana pack).
```suggestion
    "DESCRIPTION": "Remove all your debuffs. NL Draw !M! card, it costs 0 this turn. NL Gain !${ModID}:m2! ${ModID}:Portal_Sickness. NL Exhaust."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:ShowMeWhatYouGot": {
+    "NAME": "SHOW ME WHAT YOU GOT",
+    "DESCRIPTION": "Deal !D! damage to all enemies 4 times. NL Apply 3 Weak to all enemies. NL Gain 5 ${ModID}:Portal_Sickness."
```
modargo:
```suggestion
    "DESCRIPTION": "Deal !D! damage to ALL enemies 4 times. NL Apply 3 Weak to ALL enemies. NL Gain 5 ${ModID}:Portal_Sickness."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:BootlegPortalFluid": {
+    "NAME": "Bootleg Portal Fluid",
+    "DESCRIPTION": "At the start of your turn, gain 1 Strength for each distinct debuff you have. NL Gain 1 additional Strength for each debuff with more than 5 stacks."
+  },
+  "${ModID}:ParmeezianDimension": {
+    "NAME": "Parmeezian Dimension",
+    "DESCRIPTION": "Gain !B! Block. NL Transmute 1 card in your hand, reduce its cost to 0 this turn."
```
modargo:
```suggestion
    "DESCRIPTION": "Gain !B! Block. NL ${ModID}:Transmute 1 card in your hand, reduce its cost to 0 this turn."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:RoyALifeWellLived": {
+    "NAME": "Roy: A Life Well Lived",
+    "DESCRIPTION": "anniv5:transmute all cards in your hand into cards from a random pack in your pool.",
```
modargo: This should use ${ModID} instead of hardcoding the ID, and I'd suggest using "run" rather than "pool" here.
```suggestion
    "DESCRIPTION": "${ModID}:Transmute all cards in your hand into cards from a random pack in your run.",
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+    "UPGRADE_DESCRIPTION": "anniv5:transmute all cards in your hand into cards from a random pack in your pool. NL Retain."
```
modargo: Same changes for the upgraded description as for the base description, plus Retain is one of the keywords that always goes before the rest of the card text.
```suggestion
    "UPGRADE_DESCRIPTION": "Retain. NL ${ModID}:Transmute all cards in your hand into cards from a random pack in your run."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:PortalGun": {
+    "NAME": "Portal Gun",
+    "DESCRIPTION": "Deal !D! damage. NL If you have ${ModID}:Portal_Sickness, gain a random 0 cost attack into your hand."
```
modargo: The base game uses "add" rather than "gain" for things like this. 
```suggestion
    "DESCRIPTION": "Deal !D! damage. NL If you have ${ModID}:Portal_Sickness, add a random 0 cost attack into your hand."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:MortysAdventureCard": {
+    "NAME": "Morty's Adventure Card",
+    "DESCRIPTION": "Gain !${ModID}:m2! ${ModID}:Portal_Sickness. NL Choose one: NL Gain !B! Block now. NL Gain !M! Block next turn."
```
modargo: Cards like Wish separate the options with commas rather than new lines.
```suggestion
    "DESCRIPTION": "Gain !${ModID}:m2! ${ModID}:Portal_Sickness. NL Choose one: NL Gain !B! Block now, or  gain !M! Block next turn."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:PickleRick": {
+    "NAME": "Pickle Rick",
+    "DESCRIPTION": "Deal !D! damage. NL Gain !M! ${ModID}:Portal_Sickness"
```
modargo:
```suggestion
    "DESCRIPTION": "Deal !D! damage. NL Gain !M! ${ModID}:Portal_Sickness."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:WubbaLubbaDubDub": {
+    "NAME": "Wubba Lubba Dub Dub",
+    "DESCRIPTION": "Apply !M! Vulnerable and !M! Weak. NL Gain !${ModID}:m2! ${ModID}:Portal_Sickness. NL Exhaust."
+  },
+  "${ModID}:MortyGainBlockNowCard": {
+    "NAME": "Gain Block Now",
+    "DESCRIPTION": "Gain !B! block now."
```
modargo: Block should be capitalized.
```suggestion
    "DESCRIPTION": "Gain !B! Block now."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:MortyGainBlockLaterCard": {
+    "NAME": "Gain Block Later",
+    "DESCRIPTION": "Gain !B! block next turn."
```
modargo:
```suggestion
    "DESCRIPTION": "Gain !B! Block next turn."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Keywordstrings.json
```diff
@@ -0,0 +1,22 @@
...
+    "NAMES": [
+      "Portal Sickness",
+      "portal sickness",
+      "Portal_Sickness",
+      "portal_sickness"
+    ],
+    "DESCRIPTION": "Does damage to you in a number of turns. NL Additional stacks will add their damage to the same instance."
+  },
+  {
+    "PROPER_NAME": "Bootleg Portal Fluid",
```
modargo: Is this keyword needed? The card and power don't use it, as far as I can see.

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Powerstrings.json
```diff
@@ -0,0 +1,15 @@
...
+    "NAME": "Portal Sickness",
+    "DESCRIPTIONS": [
+      "Does #b%d damage to you in #b%d turns. NL Additional stacks will add their damage to this.",
+      "Does #b%d damage to you at the end of this turn. NL Additional stacks will add their damage to this."
+    ]
+  },
+  "${ModID}:BootlegPortalFluidPower": {
+    "NAME": "Bootleg Portal Fluid",
+    "DESCRIPTIONS": [
+      "At the start of your turn, gain 1 Strength for each distinct debuff you have. NL Gain 1 additional Strength for each debuff with more than 5 stacks."
```
modargo: Numbers in power descriptions are normally colored blue. It helps them stand out so you can see them at a glance.
```suggestion
      "At the start of your turn, gain #b1 Strength for each distinct debuff you have. NL Gain #b1 additional Strength for each debuff with more than #b5 stacks."
```

src/main/java/thePackmaster/actions/rickandmorty/AddPortalSicknessAction.java
```diff
@@ -0,0 +1,25 @@
...
+public class AddPortalSicknessAction extends AbstractGameAction {
+    public AddPortalSicknessAction(int amount) {
+        this.amount = amount;
+    }
+
+    @Override
+    public void update() {
+        AbstractPlayer player = Wiz.p();
+        if (!player.hasPower(PortalSicknessPower.ID)) {
+            Wiz.applyToSelf(new PortalSicknessPower(this.amount));
```
modargo: Have you tried a version of this that is only the `Wiz.applyToSelf` call here? By default, I think the turn count of the existing power will stay the same, which should be what you want. That would mean you can get rid of the hasPower check and the else branch entirely.

I mention this because right now, the code in the else branch has an issue: it won't trigger any logic for debuffs being applied to the player. So it won't be blocked by artifact, and it won't trigger debuff-synergy powers. The best way to fix those issues is to get everything to work through a normal `Wiz.applyToSelf` call. I hope it's as simple as I said above, but even if it isn't, it's worth figuring out how to do it so that artifact/debuff synergy/etc. works properly.

Once you make that change, you might not need `AddPortalSicknessAction` at all -- you could just make the `Wiz.applyToSelf` call in each spot.

src/main/java/thePackmaster/cards/rickandmortypack/BootlegPortalFluid.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+        Wiz.applyToSelf(new BootlegPortalFluidPower());
+    }
+
+    @Override
+    public void upp() {
+        upgradeBaseCost(1);
+        this.upgraded = true;
```
modargo: You don't this line; the base card class (`AbstractPackmasterCard`) handles this for you.

src/main/java/thePackmaster/cards/rickandmortypack/KeepSummerSafe.java
```diff
@@ -0,0 +1,42 @@
...
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+        Wiz.doBlk(this);
+        if (PlayerDebuffUtil.playerHasDebuff()) {
+            Wiz.atb(new DrawCardAction(2));
+        }
+    }
+
+    @Override
+    public void upp() {
+        this.upgradeBlock(6);
```
modargo: Looking at this now, block going up by more than 50% seems like too much. Going up by that much makes this one of the most powerful upgrades ever to exist; normally 1-cost block cards without a downside go up by 3 block, and even a card like Power Through with a big downside only goes up by 5 block.

This is a rare with a secondary effect, so it can have a bigger upgrade than a typical 1-cost block card, but I think a good number for this would be 4. That's still going to be a great upgrade, since you'll be happy to play this card for the mix of block and draw whenever you're debuffed.

src/main/java/thePackmaster/cards/rickandmortypack/MortyGainBlockLaterCard.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.megacrit.cardcrawl.actions.common.ApplyPowerAction;
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.powers.NextTurnBlockPower;
+import thePackmaster.util.Wiz;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class MortyGainBlockLaterCard extends AbstractRickAndMortyCard {
```
modargo, line 12: For cards like this that aren't real cards (i.e. only used for "choose one" effects or some other kind of placeholder), you should put the `@AutoAdd.Ignore` annotation above the class so that game doesn't try to add these like normal cards. It reduces confusion and avoids cluttering up the compendium even more than it already is.

src/main/java/thePackmaster/cards/rickandmortypack/MortyGainBlockNowCard.java
```diff
@@ -0,0 +1,38 @@
+package thePackmaster.cards.rickandmortypack;
+
+import com.megacrit.cardcrawl.characters.AbstractPlayer;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import thePackmaster.util.Wiz;
+
+import static thePackmaster.SpireAnniversary5Mod.makeID;
+
+public class MortyGainBlockNowCard extends AbstractRickAndMortyCard {
```
modargo, line 11: This should also have `@AutoAdd.Ignore`.

src/main/java/thePackmaster/cards/rickandmortypack/MortysMindBlowers.java
```diff
@@ -0,0 +1,47 @@
...
+    public MortysMindBlowers() {
+        super(ID, 1, CardType.SKILL, CardRarity.RARE, CardTarget.SELF);
+        this.baseMagicNumber = this.magicNumber = 1;
+        this.baseSecondMagic = this.secondMagic = 10;
+        this.exhaust = true;
+    }
+
+    @Override
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+        abstractPlayer.powers.stream()
```
modargo: You can do this more easily with `RemoveDebuffsAction` (which is what Orange Pellets and The Hierophant from the Arcana pack use):
```
Wiz.atb(new RemoveDebuffsAction(p));
```

src/main/java/thePackmaster/cards/rickandmortypack/MortysMindBlowers.java
```diff
@@ -0,0 +1,47 @@
...
+                .filter(PlayerDebuffUtil.POWER_IS_DEBUFF_PREDICATE)
+                .forEach(power -> Wiz.atb(new RemoveSpecificPowerAction(abstractPlayer, abstractPlayer, power)));
+        for (int i = 0; i < this.magicNumber; i++) {
```
modargo: This would have issues if `magicNumber` was ever greater than 1, because the top card doesn't change until the `DrawSpecificCardAction` executes, so this would add a bunch of actions trying to draw the same card.

A better way to do this is to use `DrawCardAction` with the constructor that takes an `action` parameter that gets run after the draw happens. Inside that action, you can reference `DrawCardAction.drawnCards` to get the list of cards that were drawn, and set each of them to 0-cost. That's the best way to handle effects like this.

For an example of this in action, see the code for Rummage (the Packmaster starter card). It has a very similar effect, except reducing the cost by 1 instead of setting the cost to 0.

src/main/java/thePackmaster/cards/rickandmortypack/ParmeezianDimension.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+        Wiz.doBlk(this);
+        Wiz.atb(new TransmuteCardAction((oldCard, newCard) -> newCard.setCostForTurn(0)));
+    }
+
+    @Override
+    public void upp() {
+        this.upgradeBlock(4);
```
modargo: Let's change this upgrade to give +3 block instead. That's generally what 1-cost skills without downsides get for block upgrades.

src/main/java/thePackmaster/cards/rickandmortypack/PortalGun.java
modargo: It's unlikely to happen in any normal run so that seems fine to me.

src/main/java/thePackmaster/cards/rickandmortypack/RoyALifeWellLived.java
```diff
@@ -0,0 +1,41 @@
...
+    public void use(AbstractPlayer abstractPlayer, AbstractMonster abstractMonster) {
+        AbstractCardPack newTargetPack = SpireAnniversary5Mod.getRandomPackFromCurrentPool();
+        Function<AbstractCard, Boolean> cardInNewTargetPack = card -> newTargetPack.packID.equals(Wiz.getPackByCard(card).packID);
+        Wiz.atb(new TransmuteCardAction(Integer.MAX_VALUE, false, null, cardInNewTargetPack));
+    }
+
+    @Override
+    public void upp() {
+        this.selfRetain = true;
+        this.upgrade();
```
modargo: You shouldn't need this call.

src/main/java/thePackmaster/packs/RickAndMortyPack.java
modargo: They're only a guide and we can adjust them later, so we just eyeball things for new packs. Having a 4 in frontload makes sense given all the cards that give an immediate benefit for a future price.

I'd suggest changing support to only be a 3; compared to other packs that have more in the way of draw/energy generation/etc., this pack doesn't stand out.

Otherwise the numbers seem fine.

src/main/java/thePackmaster/powers/rickandmortypack/BootlegPortalFluidPower.java
```diff
@@ -0,0 +1,40 @@
...
+    public static final String DESCRIPTION = POWER_STRINGS.DESCRIPTIONS[0];
+
+    public BootlegPortalFluidPower() {
+        super(ID, NAME, PowerType.BUFF, true, Wiz.p(), 1);
+        this.description = DESCRIPTION;
+    }
+
+    @Override
+    public void atStartOfTurn() {
+        final AtomicInteger strengthToGive = new AtomicInteger(0);
```
modargo: You can do this without needing `AtomicInteger` by doing a `map` and `sum` instead of a `forEach`. Something along these lines (haven't syntax checked this, but it should be roughly correct):
```
.mapToInt(power -> 1 + (power.amount > 5 ? 1 : 0)).sum().map(power -> 1 + (power.amount > 5 ? 1 : 0)).sum()
```

The result of that call will be a number that you can pass in to `StrengthPower`.

modargo (changes requested):
I've taken a first pass through the PR with comments covering balance, wording, and code issues -- nothing too big, just just a variety of things to fix. Hopefully shouldn't take too long, and I'll take a second pass through the review once the changes have been made.

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
grantw99: feel dumb about this one, totally thought this mod id was different but I see in the pom it's ThisModId for the expansion pack specific mod id, great call out thank you this explains a lot actually

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Keywordstrings.json
grantw99: Oooo that's my bad, good catch. I forgot that the powerstrings was where this ended up coming from. Will remove, thanks!

src/main/java/thePackmaster/actions/rickandmorty/AddPortalSicknessAction.java
grantw99: Lol I made this way harder on myself than it needed to be, apply to self totally works and I'm just gonna replace everywhere this is being used with that, big thanks

src/main/java/thePackmaster/powers/rickandmortypack/PortalSicknessPower.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void updateDescription() {
+        this.description = this.amount2 == 1 ?
+                String.format(THIS_TURN_DESCRIPTION_TEMPLATE, this.amount) :
+                String.format(FUTURE_TURN_DESCRIPTION_TEMPLATE, this.amount, this.amount2);
+    }
+
+    public void increasePortalSicknessAmount(int amount) {
```
modargo: Now that you've simplified how Portal Sickness is applied, I think this function is unused and you can delete it.

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+    "DESCRIPTION": "Gain !B! Block. NL ${ModID}:Transmute 1 card in your hand, reduce its cost to 0 this turn."
+  },
+  "${ModID}:RoyALifeWellLived": {
+    "NAME": "Roy: A Life Well Lived",
+    "DESCRIPTION": "${ModID}:Transmute all cards in your hand into cards from a random pack in your run.",
+    "UPGRADE_DESCRIPTION": "Retain. NL ${ModID}:Transmute all cards in your hand into cards from a random pack in your run."
+  },
+  "${ModID}:PortalGun": {
+    "NAME": "Portal Gun",
+    "DESCRIPTION": "Deal !D! damage. NL If you have ${ModID}:Portal_Sickness, add a random 0 cost attack into your hand."
```
modargo: Just a small thing but Attack should be capitalized here.
```suggestion
    "DESCRIPTION": "Deal !D! damage. NL If you have ${ModID}:Portal_Sickness, add a random 0 cost Attack into your hand."
```

src/main/resources/anniv5Resources/localization/eng/rickandmortypack/Cardstrings.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "${ModID}:MortysAdventureCard": {
+    "NAME": "Morty's Adventure Card",
+    "DESCRIPTION": "Gain !${ModID}:m2! ${ModID}:Portal_Sickness. NL Choose one: NL Gain !B! Block now, or  gain !M! Block next turn."
```
modargo: Getting rid of an extra space.
```suggestion
    "DESCRIPTION": "Gain !${ModID}:m2! ${ModID}:Portal_Sickness. NL Choose one: NL Gain !B! Block now, or gain !M! Block next turn."
```

modargo (commented):
Looks good. Just some dead code to clean up and a few small text fixes to make. Once that's done, should be good to go for merging.

